PR #1, opened 2021-01-22
Initial API set up 

LaurenQurashi:
This PR creates the initial files for the project, as well as creates the initial models for the database. We have outlined these models first, as they are the absolute necessities that the system needs in order to record a monthly update.


cgsunkel (approved):

api/api/supply_chain_update/models.py
```diff
@@ -0,0 +1,69 @@
+from django.db import models
+from django.conf import settings
+import uuid
+
+
+class User(models.Model):
```
tomkingston, line 6: Do you want to use this as your Django user model? If so, you will need to set the `AUTH_USER_MODEL` setting to point here. Rather than create your own user model, you might want to just have a one to one relationship to the default one instead. https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#specifying-a-custom-user-model

Also, is this the correct app to place the user model - perhaps user model and gov department would sit better in an auth app?

api/api/supply_chain_update/models.py
```diff
@@ -0,0 +1,69 @@
...
+        related_name="strategic_actions",
+    )
+
+
+class MonthlyUpdate(models.Model):
+    IMPLEMENTATION_RAG_CHOICES = [("RED", "Red"), ("AMBER", "Amber"), ("GREEN", "Green")]
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    draft = models.BooleanField(default=True)
+    submission_date = models.DateField()
+    update = models.TextField()
```
tomkingston, line 51: `update` could potentially be a slightly confusing field name, since it is a little ambiguous (you might also expect it to be a date field or even a boolean) - it's also very similar to the name of the model, so you will end up referring to `monthly_update.update` - perhaps this would be better named content?

api/api/supply_chain_update/models.py
```diff
@@ -0,0 +1,69 @@
...
+        on_delete=models.CASCADE,
+        related_name="supply_chains",
+    )
+
+
+class StrategicAction(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    name = models.CharField(max_length=settings.CHARFIELD_MAX_LENGTH)
+    description = models.TextField()
+    archived = models.BooleanField()
```
tomkingston, line 38: Looks good - I know some people like to prefix boolean fields with `is_` to make it clear it is not a date field, but I don't think that's really a problem.

tomkingston (approved):
Looks really good - since you asked for them (and it's the very start of the project), I've added some extremely picky queries / comments! 

If you are intending to use the `user` model for django auth you will need to change it soon because it can be tricky to change it once you have created the database.

api/api/supply_chain_update/models.py
LaurenQurashi: Hey! thank you this is so useful, I think we need to go away and do some more research into staff SSO as that's going to be involved with our login stuff, so we will come back to this but thank you 🎉 😄

.gitignore
```diff
@@ -0,0 +1,5 @@
+api/env/
+api/api/__pycache__/
+api/api/supply_chain_update/__pycache__/
+api/api/supply_chain_update/migrations/__pycache__/
```
tomkingston, line 4: It would be best just to ignore all pycache folders rather than specifying each of them (and also ignore *.pyc files), I've taken this from this [boilerplate](https://github.com/github/gitignore/blob/master/Python.gitignore):
```
__pycache__/
*.py[cod]
```

api/api/supply_chain_update/models.py
```diff
@@ -0,0 +1,91 @@
...
+        "GovDepartment",
+        on_delete=models.PROTECT,
+        related_name="users",
+    )
+    USERNAME_FIELD = "sso_email_user_id"
+
+
+class GovDepartment(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    name = models.CharField(max_length=settings.CHARFIELD_MAX_LENGTH)
```
tomkingston, line 36: I think these would be best moved into a new 'accounts' app ie under `api/api/accounts/models.py` - creating these models here suggests that they are coupled to this particular app and will result in database tables named `supply_chain_update__user`  and `supply_chain_update__gov_department` - it becomes a bit tricky moving this out to another app later. I think this would fall under 'separation of concerns'.

Data Hub API has a similar issue with the user (Advisor) model defined under the company app which seems unintuitive to me.

api/api/supply_chain_update/models.py
```diff
@@ -0,0 +1,91 @@
...
+        ("RED", "Red"),
+        ("AMBER", "Amber"),
+        ("GREEN", "Green"),
+    ]
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    is_draft = models.BooleanField(default=True)
+    submission_date = models.DateField()
+    content = models.TextField()
+    implementation_rag_rating = models.CharField(
+        max_length=settings.CHARFIELD_MAX_LENGTH, choices=IMPLEMENTATION_RAG_CHOICES
```
tomkingston, line 73: I guess you could set the max length to 5 as you know the choices do not allow for anything longer than that, but that doesn't make much difference. You could even change the choices to "R", "A" and "G" and set the max_length to 1 if you wanted to save some storage space, but again, there isn't a lot of point, and having the full name does make it a bit easier to see what is going on in the db.

api/api/urls.py
```diff
@@ -0,0 +1,21 @@
...
+Function views
+    1. Add an import:  from my_app import views
+    2. Add a URL to urlpatterns:  path('', views.home, name='home')
+Class-based views
+    1. Add an import:  from other_app.views import Home
+    2. Add a URL to urlpatterns:  path('', Home.as_view(), name='home')
+Including another URLconf
+    1. Import the include() function: from django.urls import include, path
+    2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
+"""
```
tomkingston, line 15: You may want to remove some of this autogenerated docstring once you are happy with the file (developers should really know this or find it in the django docs, so it's not particularly helpful).

api/api/settings.py
```diff
@@ -0,0 +1,131 @@
...
+Django settings for api project.
+
+Generated by 'django-admin startproject' using Django 3.1.5.
+
+For more information on this file, see
+https://docs.djangoproject.com/en/3.1/topics/settings/
+
+For the full list of settings and their values, see
+https://docs.djangoproject.com/en/3.1/ref/settings/
+"""
```
tomkingston, line 11: I expect you could probably cut this out or at least reduce it - developers should really know where to find the docs

api/api/settings.py
```diff
@@ -0,0 +1,131 @@
...
+from pathlib import Path
+import environ
+import os
+
+# Build paths inside the project like this: BASE_DIR / 'subdir'.
+BASE_DIR = Path(__file__).resolve().parent.parent
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/3.1/howto/deployment/checklist/
```
tomkingston, line 22: When you are happy that this is suitable for prod you'll want to remove this comment

api/api/settings.py
```diff
@@ -0,0 +1,131 @@
...
+environ.Env.read_env(env_file=os.path.join(BASE_DIR, ".env"))
+
+env = environ.Env()
+
+SECRET_KEY = env("DJANGO_SECRET_KEY")
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = True
+
+ALLOWED_HOSTS = []
```
tomkingston, line 34: You will want to get these values from env when it is ready for deployment.

On another note, I really like how you have just one settings file using environment variables rather than a number of settings files (like in data hub api). Using just one settings file is the recommended way according to [the 12 Factor App](https://12factor.net/)

api/api/supply_chain_update/models.py
```diff
@@ -0,0 +1,91 @@
...
+from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
+import uuid
+
+
+class UserManager(BaseUserManager):
+    def create_user(self):
+        pass
+
+    def create_superuser(self):
+        pass
```
tomkingston, line 12: Looks ok for now, I assume you will fill out these methods later (they will need to save the user to the database)?

tomkingston (commented):
Good stuff - I've added a few more comments, you can probably ignore quite a few of them though!

My main concern at the moment would be with having the user model inside the `supply_chain_update` app - ideally you would decouple it and create a new `accounts` app (or equivalent) that is purely responsible for managing user information.

api/api/supply_chain_update/models.py
tomkingston: Sorry, one more thing - if you think you might want to query this field a lot, it might be worth adding `db_index=True`, but you can do that when / if needed with a migration.

api/api/supply_chain_update/models.py
mforner13: Yep :) Our next steps are to create our react app and then authentication via staff-sso. As we will be creating users using data we receive from staff-sso we thought we would leave the content of these methods until then :)

api/api/supply_chain_update/models.py
mforner13: Thanks Tom - really useful! Have updated the max_length to 5. I think this field wouldn't be queried very often, but we do definitely need to consider what models need indexing going forward!

mforner13:
@tomkingston Thanks for all your help with this! Have added the suggested changes 😃

PR #3, opened 2021-02-08
Add viewset and endpoint for User model

mforner13:
This PR creates an endpoint (`/users`) which allows CRUD operations to be carried out on the `User` model, providing the user making the request is authenticated. 

As this is the first endpoint, the PR also includes setting up of the following:
- DRF's auth token app
- Factories for the `User` and `GovDepartment` models
- Fixtures providing an authenticated test client

I have also updated the black pre-commit hook to no longer specify python version 3.8.2, but rather just 3.8, to allow it to be used with any python 3.8 version.


api/api/accounts/test/factories.py
```diff
@@ -0,0 +1,29 @@
...
+    id = uuid.uuid4()
+    name = "A Government Department"
+
+    class Meta:
+        model = "accounts.GovDepartment"
+        django_get_or_create = ("id",)
+
+
+class UserFactory(factory.django.DjangoModelFactory):
+    """ User factory """
```
ambrozic, line 17: this comment also is not necessary.

api/api/accounts/test/factories.py
```diff
@@ -0,0 +1,29 @@
...
+
+    id = uuid.uuid4()
```
ambrozic, line 19: This will make all `UserFactory` objects created from this class to have same id during runtime.

mforner13:
@ambrozic Thanks for all the great comments! :) Have updated with suggestions, let us know what you think.

ambrozic (approved):

elcct (approved):

PR #4, opened 2021-02-10
Add strategic action endpoint

mforner13:
This PR adds an endpoint allowing CRUD operations on the `StrategicAction` model. It also includes a filter on the view which allows a `supply_chain_id` query parameter to be passed to the endpoint (`/strategic-actions?supply_chain_id=XXX`) which will just return strategic actions linked to that supply chain.

The tests I have included represent what this endpoint will be used for in our first feature, which will be a user giving a monthly update for a strategic action. However, any suggestions on additional necessary tests are very welcome!

### DIFF
--- a/api/api/supply_chain_update/views.py
+++ b/api/api/supply_chain_update/views.py
@@ -1,3 +1,23 @@
-from django.shortcuts import render
+from rest_framework import viewsets
+from rest_framework.permissions import IsAuthenticated
 
-# Create your views here.
+from api.supply_chain_update.models import StrategicAction
+from api.supply_chain_update.serializers import StrategicActionSerializer
+
+
+class StrategicActionViewset(viewsets.ModelViewSet):
+    """
+    A viewset that returns Strategic Action objects.
+    """
+
+    serializer_class = StrategicActionSerializer
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        queryset = StrategicAction.objects.all()
+        supply_chain_id = self.request.query_params.get(
+            "supply_chain_id",
+        )
+        if supply_chain_id:
+            queryset = queryset.filter(supply_chain__id=supply_chain_id)
+        return queryset
--- a/api/api/supply_chain_update/serializers.py
+++ b/api/api/supply_chain_update/serializers.py
@@ -1,0 +1,17 @@
+from rest_framework import serializers
+
+from api.supply_chain_update.models import StrategicAction
+
+
+class StrategicActionSerializer(serializers.ModelSerializer):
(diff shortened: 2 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -4,4 +4,5 @@ repos:
     hooks:
       - id: black
         language_version: python3.8
+        args: [--line-length=79]
```
ambrozic, line 7: do we really need this using black?

api/api/supply_chain_update/views.py
```diff
@@ -1,3 +1,24 @@
...
+    serializer_class = StrategicActionSerializer
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        queryset = StrategicAction.objects.all()
+        supply_chain_id = self.request.query_params.get(
+            "supply_chain_id",
+            None,
+        )
+        if supply_chain_id is not None:
```
ambrozic, line 22: if `supply_chain_id` is passed as `? supply_chain_id=` this condition will pass and apply empty string to queryset filter.

ambrozic (changes requested):
i find `supply_chain_update` as an app name a bit odd :)

.pre-commit-config.yaml
elcct: Such changes also should rather go into a separate PR. Is 79 character limit practical?

api/api/supply_chain_update/views.py
```diff
@@ -1,3 +1,24 @@
...
+    """
+
+    serializer_class = StrategicActionSerializer
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        queryset = StrategicAction.objects.all()
+        supply_chain_id = self.request.query_params.get(
+            "supply_chain_id",
+            None,
```
tomkingston, line 20: `None` is the default default, so you don't actually need to specify that as the second argument

api/api/supply_chain_update/views.py
mforner13: Ah thanks for spotting! Have updated and added a test for this scenario :)

.pre-commit-config.yaml
mforner13: Both good points! Have removed :)

mforner13:
@tomkingston @ambrozic @elcct Thanks for the comments! :) Have updated. 

Also we will be changing the app name to `supply_chains` in a separate PR 👍

ambrozic (approved):

elcct (approved):

tomkingston (approved):

PR #5, opened 2021-02-11
Switch database to postgres and run in docker

mforner13:
This PR sets the default database for the API as a postgres db and allows it to be run in docker. It also includes:

- a makefile with commands to create and drop the db in docker
- a db url in the sample.env

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+setup:
+	docker-compose up -d
+
+create-db: setup
+	docker exec defend-data-capture_db_1 psql -h localhost -U postgres -c "CREATE DATABASE defend WITH OWNER postgres ENCODING 'UTF8';"
+
+drop-db: setup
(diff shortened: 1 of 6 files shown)


ambrozic (approved):

LaurenQurashi (approved):

Makefile
```diff
@@ -0,0 +1,11 @@
...
+	docker-compose up -d
+
+create-db: setup
+	docker exec defend-data-capture_db_1 psql -h localhost -U postgres -c "CREATE DATABASE defend WITH OWNER postgres ENCODING 'UTF8';"
+
+drop-db: setup
+	docker exec defend-data-capture_db_1 psql -h localhost -U postgres -c "DROP DATABASE defend"
+
+tests: setup
+	cd api; pytest
```
ambrozic, line 11: you can just do `pytest api`

LaurenQurashi (approved):

PR #6, opened 2021-02-11
Add supply chain end point

LaurenQurashi:
This PR adds an endpoint with CRUD operations on the SupplyChain model. It also includes a filter on the view which allows a `gov_department_id` parameter to be passed to the endpoint (/supply-chains?gov_department_id=XXX) which will just return supply chains linked to that government department. An annotation has also been added to return the number of strategic actions linked to a supply chain, on each returned supply chain object.

The tests included check the basics of the route. Please let me know if you have any suggestions for how I can make this code better though! 🎉

### DIFF
--- a/api/api/supply_chain_update/serializers.py
+++ b/api/api/supply_chain_update/serializers.py
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from api.supply_chain_update.models import StrategicAction
+from api.supply_chain_update.models import StrategicAction, SupplyChain
 
 
 class StrategicActionSerializer(serializers.ModelSerializer):
@@ -13,5 +13,20 @@ class StrategicActionSerializer(serializers.ModelSerializer):
             "description",
             "is_archived",
             "supply_chain",
+        ]
+        depth = 1
+
+
+class SupplyChainSerializer(serializers.ModelSerializer):
+    strategic_action_count = serializers.IntegerField(read_only=True)
+
+    class Meta:
+        model = SupplyChain
+        fields = [
+            "id",
+            "name",
+            "last_submission_date",
+            "gov_department",
+            "strategic_action_count",
         ]
         depth = 1
--- a/api/api/supply_chain_update/test/test_views.py
+++ b/api/api/supply_chain_update/test/test_views.py
@@ -3,20 +3,27 @@ import pytest
 from django.urls import reverse
 from rest_framework.test import APIClient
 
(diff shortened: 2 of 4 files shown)


api/api/supply_chain_update/serializers.py
```diff
@@ -15,3 +15,18 @@ class Meta:
...
+    class Meta:
+        model = SupplyChain
+        fields = [
+            "id",
+            "name",
+            "last_submission_date",
+            "gov_department",
+            "strategic_action_count",
+        ]
+        depth = 1
```
mforner13, line 32: Does the depth add anything here?

api/api/supply_chain_update/test/test_views.py
```diff
@@ -74,3 +79,54 @@ def test_all_strategic_actions_returned_if_query_param_empty(
...
+@pytest.mark.django_db()
+def test_get_all_supply_chains(
+    test_client_with_token,
+):
+    """
+    Test that all supply chain objects are returned when an
+    authorised request is made to the '/supply-chain' endpoint.
+    """
+    supply_chain = SupplyChainFactory()
+    num_chains = SupplyChain.objects.count()
```
ambrozic, line 93: can you create multiple models here? view could be returning only single model because bugs and test would still pass.

ambrozic (approved):

api/api/supply_chain_update/serializers.py
mforner13: Ah I forgot that the gov department is a foreign key on this model! So please ignore my comment :)

api/api/supply_chain_update/test/test_views.py
```diff
@@ -74,3 +79,55 @@ def test_all_strategic_actions_returned_if_query_param_empty(
...
+@pytest.mark.django_db()
+def test_get_all_supply_chains(
+    test_client_with_token,
+):
+    """
+    Test that all supply chain objects are returned when an
+    authorised request is made to the '/supply-chain' endpoint.
+    """
+    supply_chain1 = SupplyChainFactory()
+    SupplyChainFactory.create_batch(3)
```
mforner13, line 93: You could simplify this by just creating a batch of 4, assigning it to a variable and then testing that the first one's id is returned from the response :) e.g.

`supply_chains = SupplyChainFactory.create_batch(4)`

`assert response.data[0]["id"] == str(supply_chains[0].id)`

mforner13 (approved):

PR #10, opened 2021-02-15
Add endpoint for strategic action updates

mforner13:
This PR adds an API endpoint `strategic-action-updates` to perform CRUD operations on the `StrategicActionUpdate` model. It allows two filters which can be applied to get requests:

- Filtering by `strategic_action_id`, which will get all updates related to a strategic action
- Filtering by `is_submitted`, which when true will only return updates with a status of `submitted`, and when false will return updates with a status of either `in_progress` or `completed`. This distinction is to allow users to update non-submitted updates.

### DIFF
--- a/api/api/supply_chain_update/test/test_views.py
+++ b/api/api/supply_chain_update/test/test_views.py
@@ -200,3 +200,39 @@ def test_get_all_strategic_action_updates(
     assert response.status_code == 200
     assert len(response.data) == num_updates
     assert response.data[0]["id"] == str(updates[0].id)
+
+
+@pytest.mark.django_db()
+def test_filter_updates_by_strategic_action(
+    test_client_with_token,
+):
+    """
+    Test that when a strategic action id is given in query paramters,
+    the '/strategic-action-updates' endpoint returns a list of updates
+    related to that strategic action.
+    """
+    strategic_action = StrategicActionFactory()
+    strategic_action_update = StrategicActionUpdateFactory(
+        strategic_action=strategic_action,
+    )
+    # Create additional updates not linked to this strategic action
+    StrategicActionUpdateFactory.create_batch(5)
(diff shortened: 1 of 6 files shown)


api/api/supply_chain_update/test/test_views.py
```diff
@@ -16,6 +21,7 @@
     (
         ("supply-chain-list"),
         ("strategic-action-list"),
+        ("strategic-action-update-list"),
```
ambrozic, line 24: parentheses around this strings above are redundant

api/api/supply_chain_update/views.py
```diff
@@ -45,3 +50,28 @@ def get_queryset(self):
...
+    def get_queryset(self):
+        queryset = StrategicActionUpdate.objects.all()
+
+        strategic_action_id = self.request.query_params.get("strategic_action_id")
+        if strategic_action_id:
+            queryset = queryset.filter(strategic_action__id=strategic_action_id)
+
+        is_submitted = self.request.query_params.get("is_submitted")
+        if is_submitted:
+            is_submitted = eval(is_submitted.title())
```
ambrozic, line 72: we should not be using `eval` in production. [eval executes code](https://docs.python.org/3/library/functions.html?highlight=eval#eval). lets find better way to evaluate query strings for boolean values.

ambrozic (changes requested):

api/api/supply_chain_update/views.py
mforner13: Ahhh I didn't know that, thanks for spotting! Will look into other ways :)

mforner13: Updated to use `strtobool` :) Has allowed the logic to be simplified too! Let us know what you think.

ambrozic (approved):

PR #13, opened 2021-02-24
Add email_domains field to GovDepartment model

mforner13:
This PR adds a field `email_domains` to the Govdepartment model. This field will hold a list of email domains linked to this government department, and will be used to determine what government department a user belongs to when creating a user object from an sso profile.

The PR also includes squashing previous commits and updating fixtures to contain `email_domains`.


ambrozic (approved):

ambrozic:
You could add more than a single email to _email_domains_ in factories and fixtures to make it a bit more realistic.

tomkingston (approved):

api/api/accounts/models.py
```diff
@@ -32,6 +34,27 @@ class User(AbstractBaseUser, PermissionsMixin):
     USERNAME_FIELD = "sso_email_user_id"
 
 
+def _check_email_domains_unique(email_domains):
+    for email in email_domains:
+        if GovDepartment.objects.filter(email_domains__contains=[email]):
```
ambrozic, line 39: This needs case sensitive filter. I would also prefer if all emails were saved in lower case and this functionality was tested.

ambrozic (changes requested):

api/api/accounts/models.py
```diff
@@ -33,7 +34,27 @@ class User(AbstractBaseUser, PermissionsMixin):
...
+    def __str__(self):
+        return self.name
+
+    def clean_fields(self, exclude=None):
+        super().clean_fields(exclude=exclude)
+        _check_email_domains_unique(self.email_domains)
+
+    def save(self, *args, **kwargs):
+        self.clean_fields()
+        super(GovDepartment, self).save(*args, **kwargs)
```
tomkingston, line 60: I do see the need for this here.... however, it will mean that `clean_fields` will be called twice whenever you update the GovDepartment through a ModelForm for example (once as part of the ModelForm validation and once in the save method). 
There probably aren't any significant issues with running it twice, but I think that it might be better to run the validation in the code that is creating / updating the gov department.

api/api/accounts/models.py
```diff
@@ -33,7 +34,27 @@ class User(AbstractBaseUser, PermissionsMixin):
     USERNAME_FIELD = "sso_email_user_id"
 
 
+def _check_email_domains_unique(email_domains):
+    for email in email_domains:
+        if GovDepartment.objects.filter(email_domains__contains=[email]):
+            raise ValidationError(
+                f"The domain '{email}' already exists for another government department.",
+            )
```
tomkingston, line 42: Note that if more than one of the email domains already exist, you will only raise the validation error for the first one.

tomkingston (approved):

ambrozic (approved):

api/api/accounts/models.py
mforner13: Good point Tom! As for MVP government departments will only be added via django admin, we will add in this validation when we build the ModelForm for django admin in a few weeks.

api/api/accounts/models.py
mforner13: Good point! We will add this in when adding the validation to the ModelForm :)

PR #14, opened 2021-02-26
Add authentication via staff-sso

mforner13:
This PR uses the [django-staff-sso-client](https://github.com/uktrade/django-staff-sso-client) library to authenticate users via DIT's Staff SSO service. The client will be responsible for adding users to the database using their details from staff sso. 

Because a user cannot be added without a `gov_department_id`, this PR includes a custom `create_user_mapping` method which overrides the one in the client. This method gets the user's email domain, and finds a government department with that matching domain. This logic is reliant on a government department existing with a matching email domain, in order for the user to access the app, however we think this is appropriate as there is a pre-determined group of government departments who will be accessing the tool. 

We will also include this rule in our onboarding documentation in case in the future new departments are brought into the DEFEND programme.

We were advised by webops that connecting your local application to production SSO is the best option, as it will most accurately mirror what will happen in production. In order for this to work, our callback URL (`http://localhost:8000/auth/callback/`) has been whitelisted in production SSO, and the relevant `AUTHBROKER_CLIENT_ID` and `AUTHBROKER_CLIENT_SECRET` can be found in passman.

We have added a DIT department fixture as part of this PR, so that devs working on the app locally can log in via SSO.


tomkingston (commented):
Just a few comments on the gov department validation where the email is not related to one of the gov departments

tomkingston (approved):

elcct (approved):

PR #15, opened 2021-03-01
Update project structure and naming 

LaurenQurashi:
This PR includes some renames and restructuring of the project, now that it's going to be a Django exclusive app.  This includes:
-  changing the project name from `api` to `defend_data_capture` to match the repo name
-  changing the `supply_chains_update` app to `supply_chains`
- renaming the API views and tests to reflect that they are for the API part of the Django app. 
- Moving the accounts and supply chains apps up a level in the file structure to match the traditional Django project layout.

This PR is required, as due to us converting it into a Django app, we needed to create a clear separation between the API and the rest of the app. We also don't want to explicitly refer to supply chain updates as this app could cover more than just updates in future development. 

This PR must be merged in before the PR containing the implementation of Staff SSO.


tomkingston (commented):
Hi Lauren, this looks like it will work, however, the standard django project layout would have the `accounts` and `supply_chains` apps one level up from here, which would be:

```
|- defend-data-capture
   |-app
      |- manage.py
      |- app
      |   |- __init__.py
      |   |- settings.py
      |   |- urls.py
      |   |- asgi.py
      |   |- wsgi.py
      |- accounts
      |   |- __init__.py
      |   |- models.py
      |- supply_chains
          |- __init__.py
          |- serializers.py
          |- models.py
```

This would mean you would not need to prefix all your imports from those apps, so instead of `from app.accounts.serializers import UserSerilalizer` you would just need `from accounts.serializers import UserSerilalizer`.

I think app is probably a good name (it's quite generic and lets you change things later a bit more easily), but again, this would normally be the name of the app ie `defend_data_capture`

tomkingston (approved):

PR #17, opened 2021-03-19
Amend Supply Chain model, and add additional models

LaurenQurashi:
This PR amends existing and creates new models to reflect recent changes made to the processes within the DEFEND team. 

In this PR, the following models are added:
- VulnerabilityAssessment model - responsible for holding the vulnerability assessments given by users for supply chains.
- MaturitySelfAssessment model - responsible for holding the maturity self-assessments given by users for supply chains.
- ScenarioAssessment model - responsible for holding the scenario assessments given by users for supply chains. 

The SupplyChain's model, factory, and fixtures have also been updated to include wider supply chain data. 

These changes have been made so that the data scientists knew what the shape of the data would be, but they're not added to fixtures/factories yet as no code interacts with them at the moment.


defend_data_capture/supply_chains/models.py
```diff
@@ -6,7 +6,18 @@
...
 
 
+class RAGRating(models.TextChoices):
+    RED = ("RED", "Red")
+    AMBER = ("AMBER", "Amber")
+    GREEN = ("GREEN", "Green")
+
+
 class SupplyChain(models.Model):
+    class OverallStatusRatings(models.TextChoices):
```
mforner13, line 16: I think the convention is to use the singular for classnames so `OveralStatusRating`

defend_data_capture/supply_chains/models.py
```diff
@@ -15,6 +26,20 @@ class SupplyChain(models.Model):
         on_delete=models.PROTECT,
         related_name="supply_chains",
     )
+    supply_chain_contact_name = models.CharField(
+        max_length=settings.CHARFIELD_MAX_LENGTH
+    )
+    supply_chain_contact_email = models.EmailField(blank=True)
```
mforner13, line 32: You shouldn't need to prefix these fields with supply_chain since that context comes from the model name anyway :)

defend_data_capture/supply_chains/models.py
```diff
@@ -15,6 +26,20 @@ class SupplyChain(models.Model):
...
+    overall_vulnerability_status = models.CharField(
+        choices=OverallStatusRatings.choices,
+        max_length=6,
+    )
+    disagree_overall_vulnerability_status = models.TextField(blank=True)
+    overall_risk_severity_status = models.CharField(
```
mforner13, line 38: If there are no other vulnerability or risk severity statuses within this model, I'd say you don't need the `overall` before

defend_data_capture/supply_chains/models.py
```diff
@@ -15,6 +26,20 @@ class SupplyChain(models.Model):
...
     )
+    supply_chain_contact_name = models.CharField(
+        max_length=settings.CHARFIELD_MAX_LENGTH
+    )
+    supply_chain_contact_email = models.EmailField(blank=True)
+    overall_vulnerability_status = models.CharField(
+        choices=OverallStatusRatings.choices,
+        max_length=6,
+    )
+    disagree_overall_vulnerability_status = models.TextField(blank=True)
```
mforner13, line 37: For these two disagree fields, their current name implies it would be a boolean. If these are the fields to capture the reason why they disagree I'd say have reason in the field name :)

defend_data_capture/supply_chains/models.py
mforner13: I see we used the django `EmailField` for the user model - should we do the same here? It'll say exactly what it does in the docs, but I think it provides some built-in validation :)

defend_data_capture/supply_chains/models.py
```diff
@@ -117,3 +137,113 @@ class RAGRating(models.TextChoices):
...
+    class RatingLevel(models.TextChoices):
+        LEVEL_1 = ("level_1", "Level 1")
+        LEVEL_2 = ("level_2", "Level 2")
+        LEVEL_3 = ("level_3", "Level 3")
+        LEVEL_4 = ("level_4", "Level 4")
+        LEVEL_5 = ("level_5", "Level 5")
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    maturity_rating_reason = models.TextField()
+    maturity_rating = models.CharField(
```
mforner13, line 153: Calling this `maturity_rating_level` could be a bit clearer and more in-line with the options.

defend_data_capture/supply_chains/models.py
```diff
@@ -117,3 +137,113 @@ class RAGRating(models.TextChoices):
...
+        choices=RatingLevel.choices,
+        max_length=7,
+    )
+    supply_chain = models.ForeignKey(
+        SupplyChain,
+        on_delete=models.PROTECT,
```
mforner13, line 159: If a supply chain got deleted do you think we'd want to keep the maturity assessments around? Might be a question for Shan

defend_data_capture/supply_chains/models.py
```diff
@@ -117,3 +137,113 @@ class RAGRating(models.TextChoices):
...
+    )
+    deliver = models.TextField()
+    deliver_rag_rating = models.CharField(
+        max_length=5,
+        choices=RAGRating.choices,
+        blank=True,
+    )
+    supply_chain = models.ForeignKey(
+        SupplyChain,
+        on_delete=models.PROTECT,
```
mforner13, line 200: Same question as for maturity assessment :)

defend_data_capture/supply_chains/models.py
```diff
@@ -117,3 +137,113 @@ class RAGRating(models.TextChoices):
...
+        related_name="vulnerability_assessment",
+    )
+
+
+@reversion.register()
+class ScenarioAssessment(models.Model):
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    borders_closed_rationale = models.TextField()
```
mforner13, line 209: Some help text on these rationale fields could be useful? :) I'm not quite sure what they mean -is it the rationale for the borders closed rag rating being the way it is?

mforner13 (commented):
Looks good! Nice work rationalising everything! 👍 😄  Just some comments mainly on field names.

defend_data_capture/supply_chains/models.py
```diff
@@ -117,3 +135,146 @@ class RAGRating(models.TextChoices):
...
+        on_delete=models.PROTECT,
+        related_name="vulnerability_assessment",
+    )
+
+
+@reversion.register()
+class ScenarioAssessment(models.Model):
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    borders_closed_reasoning = models.TextField(
```
mforner13, line 222: Using impact rather than reasoning could be clearer for these fields, if they're capturing the impacts of these happening? :)

mforner13 (approved):

Issue #21, opened 2021-03-31
Some of the makefile targets fail


dev-digital:
**Description**

Some of the targets within makefile(_setup-db, drop-db_) assume name of the cloned folder to be _defend_data_capture_. This assumption fail while referring name of the container for docker related commands

```docker exec $nameOfContainer```

and result with error message like 

_Error: No such container: defend-data-capture_db_1_

**Expected Behaviour**

Makefile target should execute successfully.

PR #22, opened 2021-03-31
#21: Using docker-compose to refer container instance within Makefile

dev-digital:
**Description**

Fix Makefile errors on some of the targets, refer #21 

**Changes**

- Use _docker-compose_ to reference a container instead of _docker_. Docker commands require name of the container as an argument which is not a constant

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,12 +2,12 @@ setup:
 	docker-compose up -d
 
 setup-db: setup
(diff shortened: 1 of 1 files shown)


NickFitz (approved):
👍

mforner13 (approved):
Looks good 👍

PR #24, opened 2021-04-07
Update urllib3 package due to security vulnerability

mforner13:
This PR updates the dependency `urllib` as Dependabot has flagged the previous version as having a security vulnerability ([here](https://github.com/uktrade/defend-data-capture/security/dependabot)).


NickFitz (approved):
Seems reasonable 👍

PR #30, opened 2021-04-14
Add github action for running unit tests

mforner13:
### Purpose of PR
This PR adds a github action to run both the unit and functional tests against all pull requests.

It also includes some tweaks to existing commands to make them run more efficiently (and therefore the action quicker) including:
- Not saving the video which cypress records of the functional tests if they pass
- Only killing the pid process and dropping the test database table when running the functional tests locally
- Only bringing up the db container in docker apart from in when running the functional tests (which also need the mock-sso container)

I have also provided defaults for the settings which take their values from environment variables, which avoids the need to set them as blank values for running the unit tests, as they are not needed for these tests.

### To consider
I had planned to have two separate jobs - one for unit and one for functional tests - to make it clearer which was failing in the checks. However webpack is needed to run the unit tests, as the webpack-loader `load_assets` function is called when the frontend urls are called and fails without a `webpack-stats.json` file. So to avoid installing all npm and python packages twice, I have included running both tests in the same job. 

If you have any suggestions for how to split into two jobs please let me know! 😄 

### To test
Click on the `details` link on the check in this PR to see the tests being run in action (no pun intended 😉 ).

Screenshot 2021-04-14 at 15 23 44

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 setup:
-	docker-compose up -d
+	docker-compose -f docker-compose.yaml up -d
 
 setup-db: setup
 	python defend_data_capture/manage.py migrate
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -1,0 +1,31 @@
+name: run-tests
+on: [pull_request]
+jobs:
+  run-all-tests:
+    runs-on: ubuntu-latest
+    env:
+      DATABASE_URL: postgres://postgres:password@localhost:5432/defend
+      OAUTHLIB_INSECURE_TRANSPORT: 1
+      RUNNING_LOCALLY: 0
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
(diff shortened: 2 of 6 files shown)


NickFitz (approved):
Looks good to me 👍

Makefile
```diff
@@ -1,5 +1,5 @@
 setup:
-	docker-compose up -d
+	docker-compose -f docker-compose.yaml up -d
```
dev-digital, line 2: Not sure why it was required ?

dev-digital (approved):
I presume it would notify the author on failed build ?

Makefile
mforner13: When you just run `docker-compose up` it reads both the base compose file and the override file by default. But we only want the override file (which brings up the mock-sso container) when running the functional tests. So have specified just docker-compose.yaml here to avoid bringing mock-sso up unnecessarily.

([Docker compose docs](https://docs.docker.com/compose/extends/))

mforner13: And yep if you've got your github settings set to get notifications via email you will. I got an email every time my build failed yesterday (was about 30 times after lots of experimenting 😋 )

PR #32, opened 2021-04-15
Add and update fields on strategic action model

mforner13:
This PR updates the strategic action model to include two new optional fields which have been added to the design, and makes the `supporting_organisations` field optional, as there won't always be other organisations supporting a strategic action. 

Two new fields:
- `specific_related_products` - Details of specific products within the supply chain which the action applies to, if applicable.
- `other_dependencies` - Any other dependencies or requirements for completing the action.


NickFitz (approved):
Seems harmless enough 👍

dev-digital (approved):
:+1:

PR #35, opened 2021-04-26
Add archive fields to StrategicAction & SupplyChain models

mforner13:
This PR adds the following model fields:

- SupplyChain.is_archived
- SupplyChain.archived_date
- StrategicAction.archived_reason
- StrategicAction.archived_date

It also adds logic to:

1. On both models, automatically set the `archived_date` as today on save
2. on the StrategicAction model, raise a validation error if no `archived_reason` is given but `is_archived` is set to True


NickFitz (approved):
👍

defend_data_capture/supply_chains/models.py
```diff
@@ -118,10 +124,19 @@ class SupportingOrgs(models.TextChoices):
     )
     slug = models.SlugField(null=True)
 
+    def clean(self, *args, **kwargs):
+        if self.is_archived and self.archived_reason == "":
+            raise ValidationError(
+                "An archived_reason must be given when archiving a strategic action."
+            )
+        super(StrategicAction, self).clean(*args, **kwargs)
```
dev-digital, line 132: Any reason while this type-casting is required ?
Most of the times, default would work fine

defend_data_capture/supply_chains/models.py
```diff
@@ -47,11 +48,14 @@ class StatusRating(models.TextChoices):
...
     slug = models.SlugField(null=True)
+    is_archived = models.BooleanField(default=False)
+    archived_date = models.DateField(null=True, blank=True)
 
     def save(self, *args, **kwargs):
         if not self.slug:
             self.slug = slugify(self.name)
-
+        if self.is_archived and self.archived_date is None:
+            self.archived_date = timezone.now().date()
```
dev-digital, line 58: :+1: 

So it would be in DD/MM/YYYY if the server is hosted within UK ?

dev-digital (approved):

defend_data_capture/supply_chains/models.py
mforner13: So it looks to the `TIME_ZONE` setting for the timezone - and ours is set to UTC. I think this is probs sensible so that british summertime is ignored. Although currently all our date objects are just dates, no timing.

defend_data_capture/supply_chains/models.py
mforner13: Have updated to use clean_fields instead - and this calls super in the normal way :)

PR #36, opened 2021-04-27
Fix functional test script bug & failing tests

mforner13:
This PR fixes a bug with the github action which runs tests against a PR. 

Previously, the exit code from cypress (i.e the number of tests failing from the functional test suite) was not being returned from the `run_functional_tests.sh` script. This was because the status code of the the if statement was instead being returned. 

I have added the exit code from the cypress tests to a variable and then call exit with that variable at the bottom of the file. 

As a PR with failing tests cannot be merged, I have also fixed two failing tests in the `supply_chain_summary_spec.js` which had been merged into main due to the bug in the script.

**Example**
An example of the tests successfully failing can be found in this [action run](https://github.com/uktrade/defend-data-capture/actions/runs/789322236).


dev-digital (approved):

NickFitz (approved):
Looks good :-)

PR #37, opened 2021-04-27
Refactor index view to be class based

mforner13:
This PR refactors the index view - now `HomePageView` - to be class-based rather than a function, to keep consistency between the views.

I also:
- noticed I hadn't previously added a test to make sure that the annotation on the supply_chain queryset is present, so have added an extra assertion to one of the homepage unit tests. 
- reformatted the docstrings of the homepage unit tests to be in line with PEP guidelines
- Corrected the `cy.injectAxe` command in the Readme, as it incorrectly showed `cy.injectA11y()`

### DIFF
--- a/defend_data_capture/supply_chains/views.py
+++ b/defend_data_capture/supply_chains/views.py
@@ -7,7 +7,7 @@ from django.http import HttpResponse
 from django.template import loader
 from django.db.models import Count, QuerySet
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
-from django.views.generic import TemplateView
+from django.views.generic import ListView, TemplateView
 from django.shortcuts import redirect, render
 
 from supply_chains.models import SupplyChain, StrategicAction, StrategicActionUpdate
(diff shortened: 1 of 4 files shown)


defend_data_capture/supply_chains/views.py
```diff
@@ -21,38 +21,30 @@
...
-    supply_chains = request.user.gov_department.supply_chains.order_by("name")
-    all_supply_chains = supply_chains.annotate(
-        strategic_action_count=Count("strategic_actions")
-    )
-
-    deadline = get_last_working_day_of_a_month(get_last_day_of_this_month())
-    last_deadline = get_last_working_day_of_previous_month()
-
-    num_updated_supply_chains = all_supply_chains.submitted_since(last_deadline).count()
+class HomePageView(LoginRequiredMixin, TemplateView, PaginationMixin):
```
NickFitz, line 24: The main View subclass should come at the end: `(LoginRequiredMixin, PaginationMixin, TemplateView)` because of that Python MRO stuff :-)

mforner13: Oo thanks! Have updated :)

NickFitz: As this page shows a list of supply chains, I'd suggest using a ListView:

```python
class HomePageView(LoginRequiredMixin, PaginationMixin, ListView):
    model = SupplyChain
    context_object_name = "supply_chains"
    paginate_by = 5
    template_name = "index.html"

    def get_queryset(self):
        return self.request.user.gov_department.supply_chains.annotate(
            strategic_action_count=Count("strategic_actions")
        ).order_by("name")
```

Then, in `get_context_data` you could access the supply chains QuerySet for the additional values you derive via `context['supply_chains']` after the `super()` call.

This could also give you pagination for free: https://docs.djangoproject.com/en/3.2/topics/pagination/#paginating-a-listview

N.B. the URLs in the template example there are done all wrong, they should be using `{% url %}` but I suppose they didn't want to risk confusing the point of the example by adding in other stuff :-)

dev-digital (approved):

defend_data_capture/supply_chains/views.py
NickFitz: 👍

mforner13: Good suggestion Nick thanks! Have updated to `ListView` 👍

NickFitz (approved):
👍

PR #38, opened 2021-04-28
RT-79: Supply Chain Summary Page

dev-digital:
**Description**

This PR bring in first part of work item [RT-79](https://uktrade.atlassian.net/browse/RT-79) by implementing landing page.

**Changes**
- New view _SCSummaryView_ created
- To help rendering, custom filter _boolean_status_ was added
- Updates makefile target to use e2e fixtures as test/dev data within postgres
- New module _utils.js_ was added to cypress namespace to allow sharing of common functions across tests

**Screenshot**

##### SC Summary landing page

image


defend_data_capture/supply_chains/templates/includes/sc_key_info_list.html
```diff
@@ -0,0 +1,56 @@
...
+          Change<span class="govuk-visually-hidden"> vul_status</span>
+        </a>
+      </dd>
+    </div>
+    <div class="govuk-summary-list__row govuk-summary-list__row--no-border">
+      <dt class="govuk-summary-list__key">
+        The current risk severity level is 'Medium'. Is this accurate?
+      </dt>
+      <dd class="govuk-summary-list__value">
+        {{ supply_chain.risk_severity_status|boolean_status:"Medium" }}
```
mforner13, line 48: Sorry I sent this in our Data vis Devs chat on Monday (12:22) but I think must have got lost among other messages.

I think the values we display for these questions would be based on whether there's a `disagree_reason` or not. The stakeholders don't want users to be able to directly update these statuses, so are asking the user if they agree/disagree with them.  E.g.
```
<dd class="govuk-summary-list__value">
  {% if supply_chain.risk_severity_status_disagree_reason %}
     No
     <br>
     {{ supply_chain.risk_severity_status_disagree_reason }}
  {% else %}
     Yes
  {% endif %}
</dd>
```

Makefile
```diff
@@ -17,7 +17,7 @@ tests:
 	pytest defend_data_capture
 
 load-data: setup
-	python defend_data_capture/manage.py loaddata defend_data_capture/fixtures.json
+	python defend_data_capture/manage.py loaddata cypress/fixtures/*.json
```
mforner13, line 20: Good thinking! Will make life much easier 👍

cypress/support/utils.js
```diff
@@ -0,0 +1,13 @@
+
+cy.forms = {
```
mforner13, line 2: Nice 👍  I'm guessing this will be used a lot for the rest of this summary page!

defend_data_capture/supply_chains/views.py
```diff
@@ -213,3 +214,14 @@ def get_context_data(self, **kwargs):
         )
         context["supply_chain"] = supply_chain
         return context
+
+
+class SCSummary(LoginRequiredMixin, TemplateView):
```
mforner13, line 219: I think we'd want the `GovDepPermission` mixin here too - as wouldn't want users from an unrelated gov department to see supply chain details :) And a test to make sure a non-related user gets a 403.

defend_data_capture/supply_chains/views.py
```diff
@@ -213,3 +214,14 @@ def get_context_data(self, **kwargs):
...
+    template_name = "sc_summary.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        sc_slug = kwargs.get("sc_slug", "DEFAULT")
+
+        context["supply_chain"] = SupplyChain.objects.filter(slug=sc_slug)[0]
```
mforner13, line 226: Now that we have the `is_archived` field on the SupplyChain model - I think we'd also want to filter by `is_archived`=False` (sorry this change happened after writing the user stories for this ticket, should have realised it would affect this!)

mforner13 (changes requested):

defend_data_capture/supply_chains/views.py
dev-digital: Ok, that's fine.

defend_data_capture/supply_chains/templates/includes/sc_key_info_list.html
dev-digital: Yes, you did. At that time, i chose the custom filter option having _vulnerability_status_ field. 
However I wasn't aware that reason text would also be rendered within same cell. 

Is _vulnerability_status_ used anywhere else then ?

mforner13 (approved):
Looks great! 👍

NickFitz (approved):
👍

PR #42, opened 2021-05-04
Feature/monthly update forms

NickFitz:
<del>Still a few things to sort out here:</del>

* <del>Reversion logging for changes to the strategic action target completion date;</del>
* <del>What to do about an incomplete update left over from an earlier month;</del>
* <del>Still a bit unclear what to do with the "Check your answers" page if they haven't completed all the necessary information;</del>
* <del>Probably a bunch of other things I can't remember at the moment…</del>


cypress/integration/monthly_update_forms_spec.js
```diff
@@ -0,0 +1,794 @@
...
+    },
+    reason: 'This is the reason the completion date changed'
+  },
+}
+
+Cypress.Cookies.debug(true);
+
+describe('Testing monthly update forms', () => {
+  context('for a strategic action', function() {
+    context('that has no target completion date', function() {
```
mforner13, line 148: I think splitting the file into 2 specs - with one for with target completion fate and one without - would make it more manageable :)

NickFitz: Yes, that makes sense; that also corresponds with the two user journeys in the design doc (where the first also splits in two, but let's try not to think about that right now).

cypress/fixtures/strategicActionUpdate.json
```diff
@@ -0,0 +1,104 @@
+[
```
mforner13, line 1: To avoid confusion - it could be best to add these fixtures to the existing `strategicActionUpdates.json` file. Or if they're separate on purpose, could reflect this is the file name? :)

defend_data_capture/defend_data_capture/urls.py
```diff
@@ -35,6 +42,41 @@
     path("api/", include(router.urls)),
     path("", HomePageView.as_view(), name="index"),
     path("<slug:sc_slug>/summary", SCSummary.as_view(), name="sc_summary"),
+    path(
+        "<slug:supply_chain_slug>/strategic-actions/",
```
mforner13, line 46: For all the URLs I think we need to be consistent with what we call the supply chain slug :) I'm easy either way do you guys have a preference? @NickFitz @dev-digital

defend_data_capture/supply_chains/forms.py
```diff
@@ -0,0 +1,583 @@
...
+    def make_field_required(self):
+        if self.field_to_make_required is not None:
+            self.fields[self.field_to_make_required].required = True
+
+    def make_field_not_required(self):
+        if self.field_to_make_required is not None:
+            self.fields[self.field_to_make_required].required = False
+
+
+class DetailFormMixin:
```
mforner13, line 30: A docstring for this mixin describing what its for would be really useful I think :)

defend_data_capture/supply_chains/models.py
```diff
@@ -15,6 +18,21 @@ class RAGRating(models.TextChoices):
...
+]
+
+RAGRatingHints = {
+    rating[0]: help_text
+    for rating, help_text in zip(
+        reversed(RAGRating.choices), reversed(RAGRatingHintsText)
+    )
+}
+
+
```
mforner13, line 35: Does the `reversed()` mean these hint texts will only apply to the `implementation_rag_rating` field?

defend_data_capture/supply_chains/models.py
```diff
@@ -187,13 +229,88 @@ class Status(models.TextChoices):
...
+            try:
+                kwargs.pop("force_insert")
+            except KeyError:
+                pass
+            self.save(*args, **kwargs)
+
+    def __str__(self):
+        return f"SAU: {self.strategic_action.name}, {self.slug}, {self.get_status_display()}"
+
+    def some_completion_date(self):
```
mforner13, line 247: I think this name could potentially be confusing - would `get_completion_date` be accurate?

defend_data_capture/supply_chains/templates/supply_chains/includes/mu-existing-completion-date.html
```diff
@@ -0,0 +1,14 @@
+{% if strategic_action_update.some_completion_date %}
+    <div class="govuk-inset-text">
+        <h2 class="govuk-heading-s">Current estimated date of completion</h2>
+        <p>{{ strategic_action_update.some_completion_date|date:"j F Y" }}</p>
+    </div>
```
mforner13, line 5: When the current date is `ongoing` you might need an if statement as I think it won't show text if there's a date filter being applied to it.

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-summary.html
```diff
@@ -0,0 +1,154 @@
...
+                    <dd class="govuk-summary-list__actions">
+                        <a class="govuk-link" href="{% url 'monthly-update-timing-edit' supply_chain_slug=strategic_action_update.strategic_action.supply_chain.slug strategic_action_slug=strategic_action_update.strategic_action.slug update_slug=strategic_action_update.slug %}">
+                            Change<span class="govuk-visually-hidden"> estimated date of completion</span>
+                        </a>
+                    </dd>
+                </div>
+            </dl>
+        {% endif %}
+    {% endwith %}
+    {# another chance to include the timing form… not sure what this is doing here #}
```
mforner13, line 62: Can this commented out block be removed now? :)

defend_data_capture/supply_chains/templates/supply_chains/temp-sa-list.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
```
mforner13, line 1: What's this template used for?

run_monthlyupdate_tests.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
mforner13, line 1: I don't think we'd want a separate script for just one file - but could change the script to accept an option of which spec files to run? :) Not sure if necessary though as can just run with cypress open and select the spec? :)

cypress/fixtures/strategicActionUpdate.json
NickFitz: > To avoid confusion - it could be best to add these fixtures to the existing `strategicActionUpdates.json` file. Or if they're separate on purpose, could reflect this is the file name? :)

Ah yes - I'd already created this one before the other one appeared :) I'll check if there are any conflicts and add my changes to the other one.

defend_data_capture/supply_chains/models.py
NickFitz: I'm using `reversed` because on the model they're defined as red, amber, green but on the form they're displayed as green, amber red :)

defend_data_capture/supply_chains/models.py
NickFitz: I called it `some_completion_date` as it's used when we want to know if there's either an existing or a new completion date, but don't care which kind it is. (I was tempted by `there_exists_a_completion_date` as if it was a proposition in predicate calculus, but thought that was a bit too long winded, and ∃d might not be a valid Python identifier.)

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-summary.html
NickFitz: Pretty sure it can :)

defend_data_capture/supply_chains/templates/supply_chains/temp-sa-list.html
NickFitz: That's left over from before there were working links on the task list page. I'll remove it :)

defend_data_capture/supply_chains/forms.py
NickFitz: Agreed :)

defend_data_capture/supply_chains/forms.py
```diff
@@ -0,0 +1,583 @@
...
+                    valid = config["form"].is_valid()
+        return valid
+
+    def save(self, commit=True):
+        for field_name, key, config in self.detail_forms:
+            # no need to commit if the instances are the same
+            # as the outer form will do that below (if commit=True)
+            # N.B. this assumes inner forms are using the same instance
+            # But sometimes maybe they aren't… best check for it
+            # If they aren't, we need to commit iff commit=True
```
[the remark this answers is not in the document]
NickFitz: iff: if and only if https://en.wikipedia.org/wiki/If_and_only_if :)

defend_data_capture/supply_chains/forms.py
```diff
@@ -0,0 +1,583 @@
...
+
+class MonthlyUpdateInfoForm(forms.ModelForm):
+    use_required_attribute = False
+    url_pattern_for_page = "monthly-update-info-edit"
+    content = forms.CharField(
+        required=True,
+        error_messages={
+            "required": "Enter details of the latest monthly update",
+            "max_length": "The latest monthly update should be no more than %(limit_value)d characters long (you have entered %(show_value)d)",
+            "min_length": "The latest monthly update should be at least %(limit_value)d characters long (you have entered %(show_value)d)",
```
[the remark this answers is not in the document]
NickFitz: I'm not sure that field has any limits anyway :) I think I added those while I was making sure all possible error keys that could be returned by the fields had meaningful messages, but those ones will never be used unless maximum or minimum lengths are added to the models at a later date.

defend_data_capture/supply_chains/forms.py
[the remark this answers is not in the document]
mforner13: Ahhhh my bad :p Had never seen that before!

defend_data_capture/supply_chains/templates/supply_chains/includes/mu-existing-completion-date.html
mforner13: I think we wouldn't to show that as a warning if it is ongoing though, as that's implying that it's a bad thing, whereas being 'ongoing' is a valid option. Equally if we're showing the existing date here because it's useful for the users, I think that means that it's also useful for them to see that it's ongoing? :)

defend_data_capture/supply_chains/forms.py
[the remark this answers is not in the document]
mforner13: Ah cool - I reckon remove it in that case :) As want to avoid any unused code being merged into main.

defend_data_capture/supply_chains/models.py
mforner13: Ahh okay! So the RAGChoices are used for quite a few fields, but these hint texts would only apply to the `implementation_rag_rating` sadly! So I think we need a way to only apply those hint texts to that field.

defend_data_capture/supply_chains/models.py
mforner13: Ahhhh yeah that's a tricky one! I always think that when the function name is like a question which you can answer yes/no to, it should return a boolean, whereas for this one it returns the date. What about `get_latest_completion_date` or just `latest_completion_date` ? :)

defend_data_capture/supply_chains/test/test_date_multiwidget.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_widget_accepts_date_as_value(self):
+        expected_year, expected_month, expected_date = (2020, 1, 31)
+        date_value = date(expected_year, expected_month, expected_date)
+        widget = DateMultiTextInputWidget()
+        values = widget.decompress(date_value)
+        assert 3 == len(values)
+        assert expected_date == values[0]
+        assert expected_month == values[1]
+        assert expected_year == values[2]
+        # widget_context = widget.get_context(name='test', value=date(2020, 1, 31), attrs={})
```
mforner13, line 20: Can this line be removed? :)

defend_data_capture/supply_chains/widgets.py
```diff
@@ -0,0 +1,143 @@
...
+        if isinstance(value, date):
+            return [value.day, value.month, value.year]
+        if isinstance(value, str):
+            year, month, day = value.split("-")
+            return [day, month, year]
+        return [None, None, None]
+
+    def value_from_datadict(self, data, files, name):
+        day, month, year = super().value_from_datadict(data, files, name)
+        return f"{year}-{month}-{day}"
```
mforner13, line 143: Is this used anywhere? :) (Apologies if I just missed it!)

defend_data_capture/supply_chains/test/test_models.py
```diff
@@ -88,3 +91,291 @@ def test_slug_init(self):
         # Act
         # Assert
         assert date.today().strftime("%m-%Y") == sau.slug
+
+    def test_slug_init_without_factory(self):
```
mforner13, line 95: If we're adding this test, I think we can remove the the one above (`test_slug_init`) as this is then just testing that the factory works. And since we're just using the factoryboy library for that, I'd say we shouldn't need to test it :)

defend_data_capture/supply_chains/test/test_models.py
```diff
@@ -88,3 +91,291 @@ def test_slug_init(self):
...
+        sau = StrategicActionUpdate.objects.create(
+            status=StrategicActionUpdate.Status.IN_PROGRESS,
+            supply_chain=strategic_action.supply_chain,
+            strategic_action=strategic_action,
+        )
+        # Assert
+        assert date.today().strftime("%m-%Y") == sau.slug
+
+
+@pytest.mark.django_db()
```
mforner13, line 108: The code on line 16 - `pytestmark = pytest.mark.django_db` - applies this decorator to all methods in the file, so that we don't need to apply it to each one :)

defend_data_capture/supply_chains/test/test_models.py
```diff
@@ -88,3 +91,291 @@ def test_slug_init(self):
...
+class TestStrategicActionUpdate:
```
mforner13, line 109: Should this class and the one above (`TestSAUModel`) be merged since they're both testing methods on the same model?

defend_data_capture/supply_chains/test/test_models.py
```diff
@@ -88,3 +91,291 @@ def test_slug_init(self):
...
+        changed_date = original_date + relativedelta(month=6)
+        self.strategic_action_update.changed_target_completion_date = changed_date
+        self.strategic_action_update.save()
+
+        expected_date = original_date
+        actual_date = self.strategic_action_update.some_completion_date()
+        assert actual_date is not None
+        assert actual_date == expected_date
+
+    def test_some_completion_date_for_neither_date_is_none(self):
```
mforner13, line 175: The name could be quite confusing, as it sounds like we're expecting neither date to be None - what about `test_some_completion_date_none_when_both_dates_none`

defend_data_capture/supply_chains/test/test_models.py
```diff
@@ -88,3 +91,291 @@ def test_slug_init(self):
...
+    def test_some_completion_date_for_existing_date(self):
+        assert self.strategic_action_update.some_completion_date() is not None
+
+    def test_some_completion_date_for_new_date_is_new_date(self):
+        original_date = (
+            self.strategic_action_update.strategic_action.target_completion_date
+        )
+        changed_date = original_date + relativedelta(month=6)
+        self.strategic_action_update.strategic_action.target_completion_date = None
+        self.strategic_action_update.strategic_action.save()
```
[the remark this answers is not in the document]
mforner13: Ah wait, is this testing when there isn't an existing date on the SA object? If so, I think it could be less confusing to not use the original date from the strategic action model in this test, and do something like this...

```
self.strategic_action_update.strategic_action.target_completion_date = None
self.strategic_action_update.strategic_action.save()

new_date = date.today() + relativedelta(month=6)
self.strategic_action_update.changed_target_completion_date = new_date
self.strategic_action_update.save()
expected_date = new_date
```

defend_data_capture/supply_chains/test/test_tasklist_view.py
```diff
@@ -204,7 +204,7 @@ def test_action_list(self, logged_in_client, tasklist_stub):
             iter(des_set)
         )
         assert route_set == {
-            f"{slugify(tasklist_stub['sc_name'])}/{slugify(tasklist_stub['sa_name'])}/new"
+            f"/{slugify(tasklist_stub['sc_name'])}/strategic-actions/{slugify(tasklist_stub['sa_name'])}/update/start/"
```
mforner13, line 207: I'm guessing can be just `f"/{slugify(tasklist_stub['sc_name'])}/slugify(tasklist_stub['sa_name'])}/update/start/"` once `strategic-actions` is removed from the URLs :)

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,371 @@ def get(self, request, *args, **kwargs):
         return render(request, self.template_name, context=kwargs)
 
 
+# @login_required
+# class MonthlyUpdate
+class StrategicActionListView(ListView):
+    model = StrategicAction
+    template_name = "supply_chains/temp-sa-list.html"
+    context_object_name = "strategic_actions"
```
mforner13, line 256: Can be removed :) (i think)

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,371 @@ def get(self, request, *args, **kwargs):
...
+
+
+class MonthlyUpdateMixin:
```
mforner13, line 259: Really like this way of passing functionality to the related views! 😄

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,371 @@ def get(self, request, *args, **kwargs):
...
+        strategic_action = self.get_strategic_action(
+            self.kwargs["strategic_action_slug"]
+        )
+        form.instance.strategic_action = strategic_action
+        form.instance.supply_chain = strategic_action.supply_chain
+        return super().form_valid(form)
+
+    # def get_success_url(self):
+    #     return super().get_success_url()
+    #
```
mforner13, line 384: Will this be used? :)

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,371 @@ def get(self, request, *args, **kwargs):
...
+
+
+class MonthlyUpdateInfoEditView(MonthlyUpdateMixin, UpdateView):
+    template_name = "supply_chains/monthly-update-info-form.html"
+    form_class = forms.MonthlyUpdateInfoForm
+
+    def get_success_url(self):
```
mforner13, line 391: Nice 😄  👍

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,371 @@ def get(self, request, *args, **kwargs):
...
+        return reverse(next_page_url, kwargs=url_kwargs)
+
+
+class MonthlyUpdateRevisedTimingEditView(MonthlyUpdateTimingEditView):
+    template_name = "supply_chains/monthly-update-revised-timing-form.html"
+    form_class = forms.MonthlyUpdateModifiedTimingForm
+
+    def get_success_url(self):
+        next_page_url = "monthly-update-summary"
+        # see form_valid() for an explanation of original_object
```
mforner13, line 470: Is this still relevant? :)

defend_data_capture/supply_chains/test/test_views.py
```diff
@@ -132,3 +144,464 @@ def test_strat_action_summary_page_pagination(logged_in_client, test_user):
     )
     assert response.status_code == 200
     assert len(response.context["strategic_actions"]) == 4
+
+
+@pytest.mark.django_db()
```
mforner13, line 149: Same with the decorator in this file - line 24 applies this decorator to all methods :)

defend_data_capture/supply_chains/test/test_views.py
```diff
@@ -132,3 +144,464 @@ def test_strat_action_summary_page_pagination(logged_in_client, test_user):
...
+            supply_chain=supply_chain
+        )
+        strategic_action.target_completion_date = None
+        strategic_action.is_ongoing = True
+        strategic_action.save()
+        strategic_action_update: StrategicActionUpdate = StrategicActionUpdateFactory(
+            strategic_action=strategic_action,
+            status=StrategicActionUpdate.Status.IN_PROGRESS,
+            supply_chain=strategic_action.supply_chain,
+            reason_for_completion_date_change="All things must pass.",
```
mforner13, line 165: That really made me chuckle 😄

defend_data_capture/supply_chains/test/test_monthly_update_views_user_journey.py
```diff
@@ -0,0 +1,332 @@
...
+        # form errors return 200
+        assert response.status_code == 200
+        outer_form: DetailFormMixin = response.context_data["form"]
+        inner_form = outer_form.detail_form_for_key(YesNoChoices.YES)
+        assert inner_form.errors is not None
+        assert "changed_target_completion_date" in inner_form.errors.keys()
+
+
+@pytest.mark.django_db()
+class TestMonthlyUpdateSummaryPage:
```
mforner13, line 310: Is there a test for the strategic action update being marked as 'completed' after this post request and that the strategic action date is updated (if date changes)? :) Apologies if it's in another file and I've missed it!

defend_data_capture/supply_chains/test/test_nested_detail_forms.py
```diff
@@ -0,0 +1,229 @@
...
+)
+from accounts.test.factories import GovDepartmentFactory
+
+from supply_chains import forms as our_forms
+
+
+# class TestModel(models.Model):
+#     inner_form_content = models.TextField(default='', blank=True)
+#     outer_form_content = models.TextField(default='', blank=True)
+#
```
mforner13, line 27: Can this be removed? :)

defend_data_capture/supply_chains/test/test_forms.py
```diff
@@ -0,0 +1,1107 @@
...
+
+        form_data = {
+            "implementation_rag_rating": RAGRating.RED,
+            f"{RAGRating.RED}-will_completion_date_change": YesNoChoices.NO,
+            f"{RAGRating.RED}-reason_for_delays": "Some reason",
+        }
+        form = MonthlyUpdateStatusForm(data=form_data, instance=strategic_action_update)
+        assert form.is_valid()
+        saved_instance = form.save()
+        assert saved_instance.reason_for_delays == ""
```
mforner13, line 355: Is the behaviour being tested here that if they select 'No', so the date won't change, their reason for delays doesn't get saved? If so, I think we do want to save it :)

mforner13: Ah unless is this referring to the reason for date change? (i.e the reason given on the revised timing page?) Rather than the 'explain issue' question?

defend_data_capture/supply_chains/test/test_forms.py
```diff
@@ -0,0 +1,1107 @@
...
+            f"{RAGRating.RED}-will_completion_date_change": YesNoChoices.NO,
+            f"{RAGRating.RED}-reason_for_delays": "Some reason",
+        }
+        form = MonthlyUpdateStatusForm(data=form_data, instance=strategic_action_update)
+        assert form.is_valid()
+        saved_instance = form.save()
+        assert saved_instance.changed_target_completion_date is None
+
+    # @pytest.mark.skip("Not sure if we need or want to do this")
+    def test_form_leaves_reason_for_completion_date_change_when_AMBER_and_will_completion_date_change_is_false(
```
mforner13, line 394: Is this if on Monday they select red, change target completion date. Then on Tuesday switch it to amber - what we do with their reason for date change?

I'd say if they can still access the revised timing page in this scenario - then yes we should keep the reason. But if once they've changed to amber they can't get to the revised timing page, we'd want to remove both the new date and reason for date change

defend_data_capture/supply_chains/test/test_forms.py
```diff
@@ -0,0 +1,1107 @@
...
+        detail_form = form.detail_form_for_key(RAGRating.RED)
+        assert bool(detail_form.errors) is not False
+        assert "reason_for_delays" in detail_form.errors.keys()
+
+
+@pytest.mark.django_db()
+class TestAmberReasonForDelayForm:
+    """Nothing to test unless the reason_for_delays field is required."""
+
+    pass
```
mforner13, line 526: Looks like this is covered in `test_form_requires_reason_for_delays_when_AMBER` :)

defend_data_capture/supply_chains/test/test_forms.py
```diff
@@ -0,0 +1,1107 @@
...
+        form = MonthlyUpdateModifiedTimingForm(
+            data=form_data, instance=self.strategic_action_update
+        )
+        assert not form.is_valid()
+        assert "reason_for_completion_date_change" in form.errors.keys()
+
+
+@pytest.mark.django_db()
+class TestMonthlyUpdateSubmissionFormGeneration:
+    """Should take all necessary steps to effect the transition from in progress to submitted."""
```
mforner13, line 812: I'm not sure this docstring applies to this test class or the one above? As the updates would be `completed` rather than `submitted` at the end :)

defend_data_capture/supply_chains/models.py
NickFitz: After the discussion yesterday, I've realised that those bits should always display the existing date/ongoing, even if the user is changing them, and I think this method was only being used to display the new date if there was one. If I'm correct, it's no longer needed; if I'm wrong and it's used elsewhere, I was thinking of changing it to `get_new_or_existing_completion_date` as then it should be abundantly clear what it does. But I suspect it'll just vanish :)

defend_data_capture/supply_chains/models.py
```diff
@@ -162,15 +199,20 @@ class Status(models.TextChoices):
...
-        choices=RAGRating.choices,
+        choices=reversed(RAGRating.choices),
         blank=True,
+        default=None,
+        null=True,
     )
     reason_for_delays = models.TextField(blank=True)
+    changed_target_completion_date = models.DateField(null=True, blank=True)
+    reason_for_completion_date_change = models.TextField(blank=True)
+    changed_is_ongoing = models.BooleanField(null=True, default=False)
```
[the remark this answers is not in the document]
NickFitz: On reflection, I think I'd like to leave it; it's the changed value that will be applied to the `is_ongoing` field of the SA, in the same way that `changed_target_completion_date` is the changed value for that field. But I could change them both to `changed_value_for_{thing}`? Not sure if that's overkill though :)

defend_data_capture/supply_chains/templates/supply_chains/includes/mu-existing-completion-date.html
NickFitz: This now shows 'Ongoing' if it's ongoing :)

defend_data_capture/supply_chains/templates/supply_chains/temp_mu_status_form.html
```diff
@@ -0,0 +1,27 @@
+{% extends 'supply_chains/monthly-update-form-base.html' %}
```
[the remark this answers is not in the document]
NickFitz: Deleted at my end, so it'll vanish in the next push.

defend_data_capture/supply_chains/test/test_date_multiwidget.py
NickFitz: Yes :)

defend_data_capture/supply_chains/widgets.py
NickFitz: That's used internally by Django - it's part of the process whereby the separate values from the day-month-year fields are converted to a date when a form is submitted, as described at https://docs.djangoproject.com/en/3.1/ref/forms/widgets/#django.forms.Widget.value_from_datadict 

In fact, it's nicked directly from an example further down that page of how to write a Date MultiWidget :)

defend_data_capture/supply_chains/test/test_models.py
NickFitz: Good point :)

defend_data_capture/supply_chains/test/test_models.py
NickFitz: Clever! Removed :)

defend_data_capture/supply_chains/test/test_models.py
NickFitz: Done :)

defend_data_capture/supply_chains/test/test_models.py
NickFitz: If I'm right about `some_completion_date` no longer being needed, I'll get rid of these tests too.

defend_data_capture/supply_chains/test/test_tasklist_view.py
NickFitz: Yes, good spot :)

defend_data_capture/supply_chains/views.py
NickFitz: yes, I got rid at the same time as the templates :)

defend_data_capture/supply_chains/views.py
NickFitz: Left over from early experiments, I reckon. I tend to finish up with these methods that just call `super()` because it's a convenient spot to put a breakpoint when I want to find out what Django does internally, then forget to remove them.

defend_data_capture/supply_chains/test/test_views.py
NickFitz: Great album - loads of songs that John and Paul wouldn't let George put on the Beatles albums ended up on there in 1970 instead :)

https://en.wikipedia.org/wiki/All_Things_Must_Pass

defend_data_capture/supply_chains/test/test_monthly_update_views_user_journey.py
NickFitz: That does need a test I think, but not in here as this is just for tests demonstrating that submitting a page results in a redirect to the correct next page under the various circumstances - 13 tests, which seems quite appropriate ;)

defend_data_capture/supply_chains/test/test_nested_detail_forms.py
NickFitz: Yes, as could several imports that weren't being used :)

defend_data_capture/supply_chains/test/test_forms.py
NickFitz: I need to double-check this, as I'm now wondering if I'm checking the right thing; the field on revised timing is `reason_for_completion_date_change` and the reason for delays shouldn't be going walkabout…

defend_data_capture/supply_chains/test/test_forms.py
NickFitz: Yes, another bit where I need to check the logic.

defend_data_capture/supply_chains/test/test_forms.py
NickFitz: I left those ones in as placeholders in case those two forms needed their own tests. I'll check and see if there should actually be some stuff moved in there, rather than on the form that contains them's tests.

defend_data_capture/supply_chains/test/test_forms.py
NickFitz: Yes, I need to update the docstrings. I wrote them at a point when I still wasn't;t sure how to do what that page does, which is why they're so vague :)

defend_data_capture/supply_chains/test/test_forms.py
NickFitz: It now removes the reason_for_completion_date_change when the Revised Timing page is no longer applicable.

defend_data_capture/supply_chains/test/test_forms.py
NickFitz: Updated the docstrings to more clearly reflect what's going on.

cypress/integration/monthly_update_forms_no_completion_date_spec.js
```diff
@@ -0,0 +1,980 @@
...
+  },
+  'revisedtiming': function(strategicAction) {
+    return `${baseUrl}/${strategicAction.supplyChainSlug}/${strategicAction.strategicActionSlug}/updates/${todaySlug}/revised-timing/`
+  },
+  'confirm': function(strategicAction) {
+    return `${baseUrl}/${strategicAction.supplyChainSlug}/${strategicAction.strategicActionSlug}/updates/${todaySlug}/confirm/`
+  },
+}
+
+describe('Testing monthly update forms', () => {
```
dev-digital, line 181: I presume it was a conscious decision to go for testing based on journey(merging all update pages) into one spec. If it was possible to split these into respective pages, it would have resulted in smaller code. Not asking to change here BTW :grin:

cypress/integration/monthly_update_forms_with_completion_date_spec.js
```diff
@@ -0,0 +1,961 @@
...
+    return `http://localhost:8001/${strategicAction.supplyChainSlug}/${strategicAction.strategicActionSlug}/updates/start/`
+  },
+  'info': function(strategicAction) {
+    return `http://localhost:8001/${strategicAction.supplyChainSlug}/${strategicAction.strategicActionSlug}/updates/${todaySlug}/info/`
+  },
+  'timing': function(strategicAction) {
+    return `http://localhost:8001/${strategicAction.supplyChainSlug}/${strategicAction.strategicActionSlug}/updates/${todaySlug}/timing/`
+  },
+  'status': function(strategicAction) {
+    return `http://localhost:8001/${strategicAction.supplyChainSlug}/${strategicAction.strategicActionSlug}/updates/${todaySlug}/delivery-status/`
```
dev-digital, line 172: do we need $baseURL here too ?

defend_data_capture/defend_data_capture/local_settings.py
```diff
@@ -0,0 +1,6 @@
+from defend_data_capture.settings import *
```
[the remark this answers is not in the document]
NickFitz: I assume so. It won't be used at all - I use it locally via `DJANGO_SETTINGS_MODULE`. But it would be handy if we could work out some way to exclude it from deployment,

defend_data_capture/requirements.txt
```diff
@@ -31,10 +31,12 @@ korean-lunar-calendar==0.2.1
 MarkupPy==1.14
 mypy-extensions==0.4.3
 nodeenv==1.5.0
+numpy==1.20.1
```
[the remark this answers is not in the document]
NickFitz: Not sure how they crept back in. I'll get rid :)

defend_data_capture/supply_chains/admin.py
```diff
@@ -1,3 +1,9 @@
 from django.contrib import admin
 
-# Register your models here.
+from accounts.models import GovDepartment
+from supply_chains.models import SupplyChain, StrategicAction, StrategicActionUpdate
+
+admin.site.register(GovDepartment)
+admin.site.register(SupplyChain)
+admin.site.register(StrategicAction)
+admin.site.register(StrategicActionUpdate)
```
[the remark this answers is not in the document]
mforner13: I'm changing this file as part of the admin ticket, so can remove this change from this PR :)

defend_data_capture/supply_chains/test/test_views.py
```diff
@@ -132,3 +144,461 @@ def test_strat_action_summary_page_pagination(logged_in_client, test_user):
...
+
+
+class TestNoCompletionDateMonthlyUpdateNavigationLinks:
+    def test_info_view_has_info_timing_status_summary_links(self):
+        supply_chain = SupplyChainFactory()
+        strategic_action = StrategicActionFactory(supply_chain=supply_chain)
+        strategic_action_update = StrategicActionUpdateFactory(
+            strategic_action=strategic_action,
+            supply_chain=strategic_action.supply_chain,
+        )
```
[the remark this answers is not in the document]
NickFitz: Moved into `setup_method`.

cypress/integration/monthly_update_forms_no_completion_date_spec.js
NickFitz: There's various things about the overall journey that need to be tested, such as the "next" page being different depending on what's selected, so it ended up like this :) It would probably have been better to have broken some of the tests out for individual pages though.

cypress/integration/monthly_update_forms_with_completion_date_spec.js
NickFitz: Good catch - I must have forgotten to fix this when I changed the other file :D

.gitignore
```diff
@@ -7,6 +7,8 @@ node_modules
 npm-debug.log*
 defend_data_capture/assets/webpack_bundles
 webpack-stats.json
+/env/
```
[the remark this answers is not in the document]
NickFitz: At the root of the repository, so functionally the same as env, which is already there - I've removed this line, and tidied up the .idea line after it (which ignores my PyCharm config folder).

defend_data_capture/supply_chains/test/test_forms.py
NickFitz: OK, that form doesn't really need its own tests - it's exercised by the tests for the form that wraps it, so I've got rid of that class.

defend_data_capture/supply_chains/admin.py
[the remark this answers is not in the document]
NickFitz: I had those in as I was using the admin to explore the data back at the start, and since then have used it to create new fixtures to avoid clashing with others. I'll remove them for now.

defend_data_capture/supply_chains/models.py
```diff
@@ -7,6 +9,7 @@
 
 from accounts.models import GovDepartment
 import uuid
+from .utils import get_last_working_day_of_previous_month
```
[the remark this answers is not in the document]
NickFitz: A lot of people argue that relative imports within an app are better, as they make it easier to reuse stuff in other projects. But as we're doing absolute imports, I'll fix it :)

defend_data_capture/supply_chains/models.py
```diff
@@ -187,13 +230,81 @@ class Status(models.TextChoices):
...
+            except KeyError:
+                pass
+            self.save(*args, **kwargs)
+
+    def __str__(self):
+        return f"SAU: {self.strategic_action.name}, {self.slug}, {self.get_status_display()}"
+
+    @property
+    def has_existing_target_completion_date(self):
+        return self.strategic_action.target_completion_date is not None
```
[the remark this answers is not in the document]
NickFitz: I prefer to be explicit about situations when I'm testing explicitly for None. In this case it's unlikely that a valid date could evaluate to false, but for other properties like a number that can be zero or a string that can be blank or a list that can be empty, I like to make it clear that it's specifically the absence of any value at all that's being checked.

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,372 @@ def get(self, request, *args, **kwargs):
...
+        return navigation_links
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["navigation_links"] = self.get_navigation_links()
+        return context
+
+
+class MonthlyUpdateInfoCreateView(MonthlyUpdateMixin, CreateView):
+    template_name = "supply_chains/monthly-update-info-form.html"
```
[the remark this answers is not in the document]
NickFitz: I tend to use hyphens for templates because they're (kind-of) HTML files, and hyphens are generally preferable in URLs.

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,372 @@ def get(self, request, *args, **kwargs):
         return render(request, self.template_name, context=kwargs)
 
 
+class MonthlyUpdateMixin:
+    model = StrategicActionUpdate
+    context_object_name = "strategic_action_update"
+    slug_url_kwarg = "update_slug"
```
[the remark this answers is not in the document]
NickFitz: Do you mean `slug_url_kwarg`? It's used internally by Django: https://docs.djangoproject.com/en/3.2/ref/class-based-views/mixins-single-object/#django.views.generic.detail.SingleObjectMixin.slug_url_kwarg

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,372 @@ def get(self, request, *args, **kwargs):
...
+        url_kwargs = {
+            "supply_chain_slug": self.object.strategic_action.supply_chain.slug,
+            "strategic_action_slug": self.object.strategic_action.slug,
+            "update_slug": self.object.slug,
+        }
+
+        navigation_links = {
+            "Info": {
+                "label": "Update information",
+                "url": reverse_lazy("monthly-update-info-edit", kwargs=url_kwargs),
```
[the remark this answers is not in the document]
NickFitz: I don't think it was; I can't actually remember why I used the lazy version there.

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,372 @@ def get(self, request, *args, **kwargs):
...
+        )
+
+    def get_success_url(self):
+        # This method needs to be implemented by the pages
+        # so as to express their rules for what the "next" page is
+        raise NotImplementedError(
+            f"get_success_url() not implemented by {self.__class__}"
+        )
+
+    def get_navigation_links(self):
```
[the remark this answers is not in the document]
NickFitz: I don't think there's any particular reason not to expose it as part of the view's public interface, even though there isn't currently anything external that sues it.

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,372 @@ def get(self, request, *args, **kwargs):
...
+        )
+        current_month_updates = strategic_action.monthly_updates.since(last_deadline)
+        if current_month_updates.exists():
+            current_month_update: StrategicActionUpdate = (
+                current_month_updates.order_by("date_created").last()
+            )
+        else:
+            current_month_update: StrategicActionUpdate = (
+                strategic_action.monthly_updates.create(
+                    status=StrategicActionUpdate.Status.IN_PROGRESS,
```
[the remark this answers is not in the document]
NickFitz: The problem is those navigation links at the top. When the user starts a new update, we have to create it there and then, as otherwise there's nothing for those links to point at. As it then exists, it makes sense that it's IN_PROGRESS. Otherwise, every page would have to be prepared to deal with the business of creating a new one and marking it as IN_PROGRESS, and be able to respond to a URL that doesn't include a slug.

If there was just a sequence of pages and you had to complete each one before going to the next, then a POST to that start page could do the business of creating the new update and moving it from NOT_STARTED to IN_PROGRESS, but when the user can click about and fill the whole thing in backwards if they really want to, it makes more sense (to my mind) to just create the thing right away and let them click :)

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,372 @@ def get(self, request, *args, **kwargs):
...
+        form.instance.supply_chain = strategic_action.supply_chain
+        return super().form_valid(form)
+
+
+class MonthlyUpdateInfoEditView(MonthlyUpdateMixin, UpdateView):
+    template_name = "supply_chains/monthly-update-info-form.html"
+    form_class = forms.MonthlyUpdateInfoForm
+
+    def post(self, request, *args, **kwargs):
+        return super().post(request, *args, **kwargs)
```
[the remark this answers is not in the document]
NickFitz: No, that's a leftover - I often have those as places to put a breakpoint so I can step into Django and work out why it isn't doing what I expected it to ;)

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,372 @@ def get(self, request, *args, **kwargs):
...
+
+    def get_success_url(self):
+        if self.object.strategic_action.target_completion_date is None:
+            next_page_url = "monthly-update-timing-edit"
+        else:
+            next_page_url = "monthly-update-status-edit"
```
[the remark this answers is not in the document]
NickFitz: Another case where I prefer to be explicit about testing that a property has no value at all.

defend_data_capture/supply_chains/templates/supply_chains/includes/gds-error-summary-list.html
```diff
@@ -0,0 +1,12 @@
+{% with form.url_for_errors as url_for_errors %}
+    {% for field_name, errors in form.errors.items %}
+        {% if field_name not in form.fields %}
+            {# this field must be in a detail form… #}
+            {% with form.detail_form_for_current_selection as form %}
+                {% include 'supply_chains/includes/gds-detail-error-summary-item.html' %}
```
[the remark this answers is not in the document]
NickFitz: I was going to be explicit about the fact that they were rendering stuff in the form prescribed by GDS, with the correct classes and so on. But then I forgot to apply the convention everywhere else.

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-info-form.html
```diff
@@ -0,0 +1,47 @@
+{% extends 'supply_chains/monthly-update-form-base.html' %}
```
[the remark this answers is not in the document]
NickFitz: Ah, I'm just using the styles from the included component for the breadcrumbs. The content section also changed colour in the design at some point - it used to be the hint colour.

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,374 @@ def get(self, request, *args, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["strategic_action"] = self.get_strategic_action()
+        return context
+
+    def form_valid(self, form):
+        strategic_action = self.get_strategic_action()
+        form.instance.strategic_action = strategic_action
+        form.instance.supply_chain = strategic_action.supply_chain
```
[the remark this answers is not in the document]
NickFitz: Ah good catch - that's dead code left over from when the view was creating the new SAU when the form was submitted, before I realised every page would have to be able to do that because of the sub navigation and moved to creating and redirecting.

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,361 @@ def get(self, request, *args, **kwargs):
...
+    def post(self, request, *args, **kwargs):
+        return super().post(request, *args, **kwargs)
+
+    def get_form(self, form_class=None):
+        form = super().get_form(form_class)
+        if "RED-will_completion_date_change" in self.request.POST:
+            self.completion_date_change_form = form.detail_form_for_key(RAGRating.RED)
+        return form
+
+    def get_form_kwargs(self):
```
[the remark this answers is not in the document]
NickFitz: `get_form_kwargs` is used internally by Django: https://docs.djangoproject.com/en/3.2/ref/class-based-views/mixins-editing/#django.views.generic.edit.FormMixin.get_form_kwargs

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-revised-timing-form.html
```diff
@@ -0,0 +1,53 @@
+{% extends 'supply_chains/monthly-update-form-base.html' %}
```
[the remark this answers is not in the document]
NickFitz: I wonder when that appeared in the designs. It makes no sense, as none of the other pages have that…

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-summary.html
```diff
@@ -0,0 +1,133 @@
...
+            </dt>
+            <dd class="govuk-summary-list__value">
+                {% if form.errors %}
+                    {% include "supply_chains/includes/summary_errors.html" %}
+                {% else %}
+                    {{ strategic_action_update.content }}
+                {% endif %}
+            </dd>
+            <dd class="govuk-summary-list__actions">
+                <a class="govuk-link" href="{% url 'monthly-update-info-edit'  supply_chain_slug=strategic_action_update.strategic_action.supply_chain.slug strategic_action_slug=strategic_action_update.strategic_action.slug update_slug=strategic_action_update.slug %}">
```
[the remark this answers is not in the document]
NickFitz: I'm going to say we're not doing that - there's no obvious mechanism by which this could be achieved using HTTP ;)

defend_data_capture/supply_chains/views.py
```diff
@@ -216,6 +248,361 @@ def get(self, request, *args, **kwargs):
...
+        kwargs = super().get_context_data(**kwargs)
+        if "form" in kwargs.keys():
+            kwargs["form"].is_valid()
+        return kwargs
+
+    def post(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        form = self.get_form()
+        if not form.is_valid():
+            return self.get(request, *args, **kwargs)
```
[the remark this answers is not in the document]
NickFitz: Yes, if the page has errors but the user still tries to submit it.

defend_data_capture/supply_chains/widgets.py
```diff
@@ -0,0 +1,143 @@
+from datetime import date
+
+from django.forms.widgets import RadioSelect, Textarea, MultiWidget, TextInput
+
+
+class GDSRadioSelect(RadioSelect):
```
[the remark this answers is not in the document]
NickFitz: They're leftovers from earlier experiments; removed.

defend_data_capture/supply_chains/test/test_models.py
```diff
@@ -1,14 +1,25 @@
...
+from reversion.models import Version
 
+from supply_chains.models import (
+    SupplyChain,
+    StrategicActionUpdate,
+    RAGRating,
+    StrategicAction,
+)
+from supply_chains.test.factories import SupplyChainFactory, StrategicActionFactory
 from supply_chains.models import SupplyChain, StrategicActionUpdate
```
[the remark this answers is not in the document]
NickFitz: Probably left over from a merge conflict.

defend_data_capture/supply_chains/widgets.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(self, attrs=None, choices=()):
+        if attrs is None:
+            attrs = {}
+        if "class" not in attrs.keys():
+            attrs["class"] = "govuk-radios__input"
+        super().__init__(attrs, choices)
+
+
+class HintedFieldMixin:
+    def __init__(self, hint=None, attrs=None, **kwargs):
```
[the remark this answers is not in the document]
NickFitz: Mixins _are_ regular classes. They're just classes that, by convention, have "Mixin" appended to their names to denote that they're useless when instantiated on their own but provide common functionality to other classes via multiple inheritance. As they're just classes, they all have constructors anyway; you can create an instance of, e.g. FormMixin:

```python
from django.views.generic.edit import FormMixin
fm = FormMixin()
fm.__dir__()
```
but the instance is not something that does anything useful. Same with these: they can't be used as-is, or not in any meaningful way, but separate out useful functionality that can be incorporated by other classes.

defend_data_capture/defend_data_capture/urls.py
```diff
@@ -47,4 +52,42 @@
         SAUReview.as_view(),
         name="update_review",
     ),
+    path("<slug:sc_slug>/summary", SCSummary.as_view(), name="sc_summary"),
+    path(
+        "<slug:supply_chain_slug>/<slug:strategic_action_slug>/updates/",
+        include(
+            [
+                path(
```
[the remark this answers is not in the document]
NickFitz: There's no view defined for that route, it's just a standard way of defining the common path prefix for the URLs included by that urlconf: https://docs.djangoproject.com/en/3.2/topics/http/urls/#including-other-urlconfs

defend_data_capture/supply_chains/test/test_monthly_update_views_user_journey.py
```diff
@@ -0,0 +1,332 @@
...
+        if with_monthly_update_url_kwarg:
+            url_kwargs["update_slug"] = monthly_update.slug
+    else:
+        monthly_update = None
+    url = reverse(url_name, kwargs=url_kwargs)
+    return strategic_action, monthly_update, url
+
+
+@pytest.mark.django_db()
+class TestMonthlyUpdateCreationView:
```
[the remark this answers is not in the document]
NickFitz: That's a TestCase rather than a view, but yes, the views need to check login status.

defend_data_capture/supply_chains/forms.py
```diff
@@ -0,0 +1,664 @@
...
+            if isinstance(fields_to_make_required, str):
+                fields = [fields_to_make_required]
+            else:
+                fields = fields_to_make_required
+            for field_name in fields:
+                if field_name in self.fields:
+                    self.fields[field_name].required = False
+
+
+class DetailFormMixin:
```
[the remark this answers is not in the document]
NickFitz: Mixin again: useless on its own, but part of the shared functionality it provides to classes that inherit from it is in the __init__ method.

defend_data_capture/supply_chains/forms.py
```diff
@@ -0,0 +1,664 @@
...
+                form_class = config["form_class"]
+                kwargs["prefix"] = key
+                config["form"] = form_class(*args, **kwargs)
+                self._detail_forms_dict[key] = config["form"]
+            except KeyError:
+                pass
+
+    def detail_form_for_current_selection(self):
+        if hasattr(self, "cleaned_data"):
+            if self.detail_selection_field:
```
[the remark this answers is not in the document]
NickFitz: Caught me :D

defend_data_capture/supply_chains/models.py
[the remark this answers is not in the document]
mforner13: I think `changed_value_for_{thing}` is a good idea :) or `new_target_completion_date` and `new_value_for_is_ongoing` - but am easy either way!

defend_data_capture/supply_chains/templates/supply_chains/includes/gds-error-summary-list.html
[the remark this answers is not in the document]
mforner13: If we're keeping the prefix - I'd rather we use `govuk-..` as that's really where it's come from (although GDS surely did create the design system :p)

defend_data_capture/supply_chains/models.py
NickFitz: I've renamed them to `ImplementationRAGRatingHints` and moved them into `forms.py` where they're actually used :)

defend_data_capture/defend_data_capture/urls.py
NickFitz: Leaving this for later, as agreed :)

defend_data_capture/supply_chains/models.py
NickFitz: It vanished :)

run_monthlyupdate_tests.sh
NickFitz: Got rid :)

defend_data_capture/supply_chains/views.py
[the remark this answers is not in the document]
NickFitz: Turns out I hadn't even been consistent myself ;) Switched them to snake-case for consistency.

defend_data_capture/supply_chains/views.py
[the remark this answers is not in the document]
NickFitz: One thing I noticed in the task list is that it doesn't seem to have consistent, or at least persistent, ordering. On one occasion, the item I wanted was on the second page, but when I went to the second page, it had moved to the first page.

defend_data_capture/supply_chains/templates/supply_chains/includes/gds-error-summary-list.html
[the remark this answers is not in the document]
NickFitz: I've removed the prefixes at the same time as snake_casing everything, as I hadn't been consistent about it anyway :D

requirements.txt
```diff
@@ -33,10 +33,12 @@ korean-lunar-calendar==0.2.1
 MarkupPy==1.14
 mypy-extensions==0.4.3
 nodeenv==1.5.0
+numpy==1.20.1
```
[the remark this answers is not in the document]
NickFitz: I got rid of them again. Let's see if they stay away this time :)

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-revised-timing-form.html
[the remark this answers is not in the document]
NickFitz: Seems like it was a mistake.

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-summary.html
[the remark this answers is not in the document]
NickFitz: Antoine has agreed to leave this for now, as it's potentially a real pain thanks to those links at the top :)

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-summary.html
```diff
@@ -0,0 +1,133 @@
...
+                        {% else %}
+                            {% if  strategic_action_update.has_new_is_ongoing%}
+                                Ongoing
+                            {% else %}
+                                {{ strategic_action_update.changed_target_completion_date|date:"j F Y" }}
+                            {% endif %}
+                        {% endif %}
+                    </dd>
+                    <dd class="govuk-summary-list__actions">
+                        <a class="govuk-link" href="{% url 'monthly-update-timing-edit' supply_chain_slug=strategic_action_update.strategic_action.supply_chain.slug strategic_action_slug=strategic_action_update.strategic_action.slug update_slug=strategic_action_update.slug %}">
```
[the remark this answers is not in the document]
NickFitz: Sadly, long lines are inevitable in Django templates due to their inability to let template tags span more than one line :)

defend_data_capture/supply_chains/models.py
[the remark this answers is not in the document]
NickFitz: Done :)

defend_data_capture/supply_chains/templates/supply_chains/monthly-update-info-form.html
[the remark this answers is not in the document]
NickFitz: Fixed :)

defend_data_capture/supply_chains/views.py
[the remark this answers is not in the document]
dev-digital: Thanks.
For my own understanding, is it used internally. in other words, would it break if I comment that line ?

defend_data_capture/supply_chains/widgets.py
```diff
@@ -0,0 +1,143 @@
...
+            ),
+        }
+        super().__init__(widgets, attrs)
+        self.labels = labels
+        self.govuk_hint = hint
+        self.legend = legend
+
+    def get_context(self, name, value, attrs):
+        context = super().get_context(name, value, attrs)
+        if self.labels is not None:
```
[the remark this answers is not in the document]
NickFitz: I've removed the occurrences of `is not None` as they do seem to be superfluous. But I've left several occurrences of `if {thing} is None` as they are in contexts where either an argument has a default of None or a value has been obtained from a dict via `{dict}.get({key}, None)`. The `is None` idiom in those contexts not only makes it explicit that the code is handling the case of providing a default value, but also ensures that a value that was actually passed (such as an empty queryset or list) which is valid input but which happens to equate to `False` isn't replaced with the default, which may not itself equate to `False`.

defend_data_capture/supply_chains/views.py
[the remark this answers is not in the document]
dev-digital: Ok. This component has given(still giving) us lot of trouble. It doesn't make sense to me having mandatory fields in the form and allow user to jump around via breadcrumbs.

For tasklist, it would sort all SA per custom order. Position of an element depends on other SAs. If its not behaving as expected, happy to look at the issue.

defend_data_capture/supply_chains/views.py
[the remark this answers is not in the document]
dev-digital: Ok, thanks.
I meant, like _post_ method, its calling the parent implementation. do we need to retain it here ?

defend_data_capture/supply_chains/views.py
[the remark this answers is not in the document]
dev-digital: Yes, for some reason I thought validation would have disabled the button.

defend_data_capture/supply_chains/test/test_monthly_update_views_user_journey.py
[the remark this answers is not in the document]
NickFitz: Added LoginRequiredMixin to all views and updated tests accordingly :)

defend_data_capture/supply_chains/test/test_models.py
NickFitz: I do it that way because the important thing is that the new date replaces the existing date, and there's a very slight chance that one day the existing date could be the same as today plus delta, which could lead to the test passing when it should fail on that particular day. By ensuring that the new date is the existing date plus some delta, it's guaranteed that the test will only pass if things are working properly. I know it's very unlikely that something would break the specific bit of code under test on that specific day, but you can't be too careful ;)

defend_data_capture/supply_chains/test/test_forms.py
mforner13: Did we decide what to do with this test? :) (can't remember)

NickFitz: That one and a few others are actually skipped :) I'll remove them so they don't clutter the place up and cause confusion.

defend_data_capture/supply_chains/forms.py
```diff
@@ -0,0 +1,664 @@
+from datetime import date
```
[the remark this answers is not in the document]
NickFitz: Agreed. I've just been through and rearranged imports as described, removed unused imports, and normalised situations such as {model}.forms being imported in the same file as {module}.forms.{class} leading to inconsistent usage - sometimes forms.{class}, sometimes {class}.

mforner13 (approved):
Looks good - pending GitHub actions coming back to life and tests passing 😄

dev-digital:
Here are few things out-standing

- Review page is broken : throws 404
http://localhost:8001/supply-chain-5/strategic-action-0/updates/05-2021/info//review

There was another 404 error I had found during initial review(I can't find it now), not sure if it got resolved

- Update related view also need to derive from PermissionMixin. (If you can), add tests for validating them ie AUTH and PERMISSION

Happy to help with tests or anything if required.

NickFitz:
For some reason GitHub won't let me comment directly. @dev-digital That URL isn't defined anywhere (and also has two slashes). The nearest one to it would be http://localhost:8001/supply-chain-5/strategic-action-0/05-2021/review which gives an `IndexError` in `SAUReview`.

I've added the `GovDepPermissionMixin` to the views but now everything fails because the mixin has the slug name hardcoded. This will ultimately be resolved by normalising the URLs and slug names, but we could also add an additional field to the views defining the `supply_chain_slug_url_kwarg` (for consistency with Django's own `slug_url_kwarg` which all views should have defined). That way, we avoid hardcoding the magic string in the mixin.

dev-digital (approved):

PR #46, opened 2021-05-17
Feature/fixture fixup

NickFitz:
This overrides the `loaddata` management command to fix the submission dates on `StrategicActionUpdate` objects after the fixtures are loaded.


defend_data_capture/supply_chains/management/commands/loaddata.py
```diff
@@ -0,0 +1,31 @@
+from datetime import date
```
dev-digital, line 1: So this PR brings in changes for RT-169 which affect testing and test fixtures only. I like the way changes are done to the (test)model. However I can't establish the need to over-ride _loaddata_ command within the source(part of deployed modules)

By the looks of it, we going to (only) need _update_submission_dates_ called after loading fixtures within _run_functional_tests.sh_

Doing this would contain changes within test env w/o causing any adverse effects.

This is how I see things, happy to discuss and agree with rest of the team

defend_data_capture/supply_chains/management/commands/loaddata.py
```diff
@@ -0,0 +1,31 @@
...
+        updates = StrategicActionUpdate.objects.all()
+        self.update_submission_dates(updates, months_to_add)
+
+    def update_submission_dates(self, updates, months_to_add):
+        updated_updates = []
+        for update in updates:
+            if update.submission_date is not None:
+                update.submission_date += months_to_add
+                updated_updates.append(update)
+        StrategicActionUpdate.objects.bulk_update(updated_updates, ["submission_date"])
```
dev-digital, line 24: I think there are other places needing updates like (just example and not exhaustive list)

SupplyChain.last_submission_date
StrategicActionUpdate.date_created

defend_data_capture/supply_chains/management/commands/loaddata.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def calculate_months_to_add(self):
```
dev-digital, line 26: Assuming we are going include all the date related fields within model and apply offset.

There could a case where we want to exclude this offset change for a particular test(include/exclude list)

defend_data_capture/supply_chains/management/commands/loaddata.py
NickFitz: It seems to me that the data installed from the fixtures also need to be updated when running tests locally or when installing fixtures for local development; doing this in `loaddata` ensures that this happens consistently.

defend_data_capture/supply_chains/management/commands/loaddata.py
NickFitz: I wasn't sure if those values would affect anything we currently have, but I'll have a look at patching them too :)

defend_data_capture/supply_chains/management/commands/loaddata.py
NickFitz: I can look into adding that if it seems likely to be useful - shouldn't be too hard to implement by listing the excluded fixture's primary key :)

defend_data_capture/supply_chains/management/commands/loaddata.py
mforner13: I think we'd only need to adjust dates which affect monthly submissions with this command tbh, as for example the start date of a strategic action doesn't affect any logic.

defend_data_capture/supply_chains/management/commands/loaddata.py
mforner13: I've been trying to think if there's a time we'd want to load data but not update the fixtures and I'm not sure there is tbh, as I'd imagine we'll always need updates of different stages, and would be useful to have these when working locally as well as in the functional tests :) So I'm pro having it in loaddata :)

mforner13 (approved):
Looks good to me 👍

defend_data_capture/supply_chains/management/commands/loaddata.py
dev-digital: True.

At the time of writing this I didn't check

> SupplyChain.last_submission_date
> StrategicActionUpdate.date_created

Now I can confirm, w/o changing above fields come June and tests will fail.

dev-digital (approved):

PR #49, opened 2021-05-18
Create django admin

mforner13:
This PR creates a django admin instance for the application.

**Configuring display of objects**
From feedback we had from future admin users on how they identify different objects, I have configured the appropriate values to appear in the list view for each model. Also, in order for the objects to be easily recognisable, I have added `__str__` methods to models which didn't have them.

**Logging into admin**
To avoid the need for admin users to have two sets of login details, I have overridden the `login()` method for the admin site. 

If a user is already authenticated (via SSO) and they are marked as `is_staff=True` they will be able to access the admin index page. If either of these are not true, they will be redirected to the app's index url, where if they're unauthenticated they will be redirected to SSO to login.

**Custom email domains field**
For the `GovDepartment.email_domains` field, as it is an `ArrayField`, I have created a form for this model, which specifies that the field type for email_domains is a [SplitArrayField](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/forms/#splitarrayfield). This provides a simpler UI for admin users adding new gov departments.

Screenshot 2021-05-18 at 14 01 36

**NB this PR also removes the `--collect-only` flag accidentally left in the GH action to run pytest.

### DIFF
--- a/defend_data_capture/defend_data_capture/settings.py
+++ b/defend_data_capture/defend_data_capture/settings.py
@@ -28,11 +28,12 @@ ALLOWED_HOSTS = env.list("ALLOWED_HOSTS", default=["localhost"])
 
 INSTALLED_APPS = [
     "authbroker_client",
-    "django.contrib.admin",
+    "django.contrib.admin.apps.SimpleAdminConfig",
     "django.contrib.auth",
     "django.contrib.contenttypes",
-    "django.contrib.sessions",
     "django.contrib.messages",
+    "django.contrib.postgres",
+    "django.contrib.sessions",
     "django.contrib.staticfiles",
     "supply_chains",
     "accounts",
--- a/defend_data_capture/supply_chains/templates/base.html
+++ b/defend_data_capture/supply_chains/templates/base.html
@@ -41,7 +41,7 @@
                         Resilience tool
                     </a>
                     <span class="app-header-span">
-                        <p class="app-header-item">{{ request.user.first_name }} {{ request.user.last_name }} - {{ request.user.gov_department.name }}</p>
+                        <p class="app-header-item">{{ request.user }} - {{ request.user.gov_department.name }}</p>
                     </span>
                 </div>
             </div>
--- a/defend_data_capture/supply_chains/test/test_admin.py
+++ b/defend_data_capture/supply_chains/test/test_admin.py
@@ -1,0 +1,43 @@
+import pytest
+
+from django.test import Client
+from django.urls import reverse
+
+
+@pytest.mark.django_db
+class TestAdminViews:
+    def test_index_redirects_to_login_unauthenticated(self):
+        """Test unauthenticated user is redirected to admin login."""
+        client = Client()
+        expected_redirect_url = reverse("admin:login") + "?next=/admin/"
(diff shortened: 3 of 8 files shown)


defend_data_capture/supply_chains/test/test_admin.py
```diff
@@ -0,0 +1,27 @@
...
+from django.urls import reverse
+
+
+@pytest.mark.django_db
+class TestAdminLogin:
+    def test_login_unauthenticated(self):
+        """Test unauthenticated user is redirected to auth login url."""
+        client = Client()
+        response = client.get(reverse("admin:index"))
+        assert response.status_code == 302
```
mforner13, line 13: I was unsure of how to check they've been redirected away from admin, since we don't want to be calling staff SSO in the tests (or have to set up mock sso just for this).

Any ideas would be v appreciated :)

NickFitz: My test_monthly_update_views_user_journey.py tests check for being redirected to a specific URL., if that's what you need. For example:
```python
        expected_redirect_url = reverse(
            "monthly-update-info-edit",
            kwargs={
                "supply_chain_slug": strategic_action.supply_chain.slug,
                "strategic_action_slug": strategic_action.slug,
                "update_slug": monthly_update.slug,
            },
        )
        assert response.status_code == 302
        assert response.url == expected_redirect_url
```

defend_data_capture/defend_data_capture/settings.py
```diff
@@ -28,11 +28,12 @@
 
 INSTALLED_APPS = [
     "authbroker_client",
-    "django.contrib.admin",
+    "django.contrib.admin.apps.SimpleAdminConfig",
     "django.contrib.auth",
     "django.contrib.contenttypes",
-    "django.contrib.sessions",
     "django.contrib.messages",
+    "django.contrib.postgres",
```
dev-digital, line 35: Why was this required,  more for my knowledge

dev-digital (approved):
Looks good to me :+1:

defend_data_capture/supply_chains/templates/base.html
```diff
@@ -41,7 +41,7 @@
                         Resilience tool
                     </a>
                     <span class="app-header-span">
-                        <p class="app-header-item">{{ request.user.first_name }} {{ request.user.last_name }} - {{ request.user.gov_department.name }}</p>
+                        <p class="app-header-item">{{ request.user }} - {{ request.user.gov_department.name }}</p>
```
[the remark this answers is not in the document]
mforner13: So now that we have the `__str__` method for the user model which shows first name + last name, in the template, we can just refer to them as `request.user` to get the first name + last name, as it gets the string version automatically :)

defend_data_capture/defend_data_capture/settings.py
mforner13: Good question! It was to allow us to use the the Postgres SpitArrayField for the gov department email_domains field :) 
`from django.contrib.postgres.forms.array import SplitArrayField`

dev-digital (approved):

NickFitz (approved):
👍 +1

PR #51, opened 2021-05-19
Tidy up URLconfigs.

NickFitz:
Apart from rearranging `urls.py` to make it easier to add more URLs in the future, the main changes are to ensure all URLs end in a / and a couple of minor changes to path components to better reflect the content or purpose of the associated resource.

### DIFF
--- a/defend_data_capture/defend_data_capture/urls.py
+++ b/defend_data_capture/defend_data_capture/urls.py
@@ -64,7 +64,7 @@ monthly_update_urlpatterns = [
     path(
         "review/",
         SAUReview.as_view(),
-        name="update_review",
+        name="monthly-update-review",
     ),
 ]
 
@@ -72,10 +72,10 @@ strategic_action_urlpatterns = [
     path(
         "strategic-actions/",
         SASummaryView.as_view(),
-        name="strat_action_summary",
+        name="strategic-action-summary",
     ),
     path(
-        "<slug:action_slug>/update/",
+        "<slug:action_slug>/updates/",
         include(
             [
                 path(
@@ -95,12 +95,12 @@ supply_chain_urlpatterns = [
         "<slug:supply_chain_slug>/",
         include(
             [
-                path("", SCTaskListView.as_view(), name="tlist"),
-                path("summary/", SCSummary.as_view(), name="sc_summary"),
+                path("", SCTaskListView.as_view(), name="supply-chain-task-list"),
+                path("summary/", SCSummary.as_view(), name="supply-chain-summary"),
                 path(
                     "complete/",
                     SCCompleteView.as_view(),
-                    name="update_complete",
+                    name="supply-chain-update-complete",
                 ),
                 path("", include(strategic_action_urlpatterns)),
             ]
--- a/defend_data_capture/defend_data_capture/urls.py
+++ b/defend_data_capture/defend_data_capture/urls.py
@@ -1,5 +1,3 @@
-from django.conf import settings
-from django.contrib import admin
 from django.urls import path, include
 from rest_framework import routers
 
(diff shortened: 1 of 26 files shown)


defend_data_capture/defend_data_capture/urls.py
```diff
@@ -37,63 +35,82 @@
...
         name="update_review",
     ),
-    path("<slug:supply_chain_slug>/summary", SCSummary.as_view(), name="sc_summary"),
+]
+
+strategic_action_urlpatterns = [
+    path(
+        "strategic-actions/",
+        SASummaryView.as_view(),
+        name="strat_action_summary",
```
dev-digital, line 75: We going to normalise name too, aren't we ?

image

defend_data_capture/defend_data_capture/urls.py
```diff
@@ -37,63 +35,82 @@
...
+    ),
     path(
-        "<slug:supply_chain_slug>/<slug:action_slug>/updates/",
+        "<slug:action_slug>/update/",
```
dev-digital, line 78: Just curious why slug/endpoint went to singular form.

defend_data_capture/defend_data_capture/urls.py
```diff
@@ -37,63 +35,82 @@
...
-    path("admin/", admin_site.urls),
-    path("api/", include(router.urls)),
-    path("", HomePageView.as_view(), name="index"),
-    path("<slug:supply_chain_slug>", SCTaskListView.as_view(), name="tlist"),
+monthly_update_urlpatterns = [
     path(
-        "<slug:supply_chain_slug>/complete",
-        SCCompleteView.as_view(),
-        name="update_complete",
+        "latest/",
```
dev-digital, line 40: Another curiosity question, why it moved from _info_ which reflected what page was doing.

defend_data_capture/supply_chains/test/test_tasklist_view.py
```diff
@@ -210,16 +210,32 @@ def test_action_list(self, logged_in_client, tasklist_stub):
             iter(des_set)
         )
         assert route_set == {
-            f"/{slugify(tasklist_stub['sc_name'])}/{slugify(tasklist_stub['sa_name'])}/updates/start/"
+            reverse(
```
dev-digital, line 213: :+1:

defend_data_capture/supply_chains/views.py
```diff
@@ -81,13 +81,20 @@ class SCTaskListView(
...
     def _update_review_routes(self) -> None:
         for update in self.sa_updates:
-            tokens = update["route"].split("/")
-
-            # Remove slug 'info' and trailing '/'
-            tokens = tokens[:-2]
-
-            tokens.append("review")
-            update["route"] = "/".join(tokens)
+            url_kwargs = {
```
dev-digital, line 84: It didn't feel right yesterday while I was _man handling_ the route but I just went through it.
Agree, its the right way. Can we get rid of comments now.

dev-digital (commented):
Minor comments w/o approval.

defend_data_capture/defend_data_capture/urls.py
NickFitz: Because in the context of the SA, the forms are for the latest update rather than a collection of updates. But I'm not too fussed if we use updates instead :)

defend_data_capture/defend_data_capture/urls.py
NickFitz: I thought of doing that but decided to hold off as it involves a lot of search-and-replace that would touch many files, and I figured it might be better to avoid that when the main thing was to sort out the way they look. But I can also address this if it seems worthwhile getting it done now.

defend_data_capture/supply_chains/views.py
```diff
@@ -81,13 +81,20 @@ class SCTaskListView(
...
+            url_kwargs.update(self.kwargs)
+            url = reverse("update_review", kwargs=url_kwargs)
+            update["route"] = url
+            # tokens = update["route"].split("/")
+            #
+            # # Remove slug 'info' and trailing '/'
+            # tokens = tokens[:-2]
+            #
+            # tokens.append("review")
+            # update["route"] = "/".join(tokens)
```
mforner13, line 97: Can this be removed? :)

defend_data_capture/defend_data_capture/urls.py
NickFitz: The h2 heading refers to it as "Latest monthly update" so I thought that reflecting the header content made sense :)

defend_data_capture/supply_chains/views.py
NickFitz: Yes, it was mainly left in for reference in case the alternative turned out not to work as expected :)

defend_data_capture/defend_data_capture/urls.py
mforner13: I think this could be quite confusing, as to me that URL implies it's the latest update given. I think info is fine tbh :)

defend_data_capture/defend_data_capture/urls.py
mforner13: Sorry this will add more to this PR, but I think as names are currently inconsistent, it would be worth making them all kebab case :)

defend_data_capture/defend_data_capture/urls.py
dev-digital: Same here.

This is what made me thing about it.

_<slug:action_slug>/**update/<slug:update_slug>**/info_

As in one among many

NickFitz: Yes, I suppose if there's a specific identifier like 05-2021 for the current update, that implies it's part of a collection of updates. I'll revert it :)

defend_data_capture/defend_data_capture/urls.py
```diff
@@ -95,12 +95,12 @@
...
             [
-                path("", SCTaskListView.as_view(), name="tlist"),
-                path("summary/", SCSummary.as_view(), name="sc_summary"),
+                path("", SCTaskListView.as_view(), name="supply-chain-task-list"),
+                path("summary/", SCSummary.as_view(), name="supply-chain-summary"),
                 path(
                     "complete/",
                     SCCompleteView.as_view(),
-                    name="update_complete",
+                    name="supply-chain-update-complete",
```
mforner13, line 103: Good thinking distinguishing the supply chain and monthly update 👍

mforner13 (approved):
Looks good - nice and neat 👍

dev-digital (approved):
Looks good :+1:

PR #53, opened 2021-05-20
RT-214: Data loader utility

dev-digital:
**Description**

This PR brings in RT-214 related changes

**Changes**
- Add _data_loader_ Django admin command to import CSV formatted data
- Tests and related data files
- Minor migration changes - Need to apply migrations on pulling down this change-set


defend_data_capture/supply_chains/management/commands/data_loader.py
```diff
@@ -0,0 +1,134 @@
...
+                )
+
+                row["fields"].pop("str( not required in database)", None)
+
+            elif row["model"] == MODEL_STRAT_ACTION_UPDATE:
+                self._update_date_fields(
+                    row["fields"], "submission_date", "date_created"
+                )
+
+                # TODO: Check with DE
```
dev-digital, line 104: Are we good with this assumption ?

defend_data_capture/supply_chains/management/commands/data_loader.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+
+    def _get_json_object(self, csv_file: str) -> object:
+        with open(csv_file) as f:
+            reader = csv.DictReader(f)
+            rows = list(reader)
+
+        return json.dumps(rows)
+
+    def _format_json_object(self, model: str, rows: object) -> List:
```
NickFitz, line 45: As `rows` in this method is a list that's iterated over, `collections.Iterable` would be the appropriate type hint rather than `object`.

defend_data_capture/supply_chains/management/commands/data_loader.py
```diff
@@ -0,0 +1,134 @@
...
+                f"Unknown model {options['model']}. \n\nRefer help for supported values"
+            )
+
+        obj = json.loads(self._get_json_object(options["csvfile"]))
+        json_obj = self._format_json_object(options["model"], obj)
+
+        with NamedTemporaryFile(suffix=".json", mode="w") as fp:
+            json.dump(json_obj, fp)
+            fp.seek(0)
+            management.call_command(loaddata.Command(), fp.name, format="json")
```
NickFitz, line 129: You could just use the name of the command here: `management.call_command("loaddata",…` But I don't think it really matters either way :)

NickFitz (approved):

defend_data_capture/supply_chains/management/commands/data_loader.py
dev-digital: Sure as I List imported already, I would use that instead.

defend_data_capture/supply_chains/management/commands/data_loader.py
mforner13: This is tricky! I'm not sure we'd want it as could be quite misleading. Could we automatically set it as the submission date instead? Assuming existing ones already have submission dates :)

dev-digital: Yes, that can do that if you going to approve it :grimacing: 

I plan to get this clarified once shan is back anyway.

defend_data_capture/supply_chains/management/commands/data_loader.py
```diff
@@ -0,0 +1,134 @@
...
+
+ALL_MODELS = [
+    MODEL_GOV_DEPT,
+    MODEL_SUPPLY_CHAIN,
+    MODEL_STRAT_ACTION,
+    MODEL_STRAT_ACTION_UPDATE,
+]
+
+
+class Command(BaseCommand):
```
mforner13, line 24: Since we're going to have the loaddata mgmt command too, to clearly show the difference perhaps this one could. be called `ingest_csv` or `upload_csv` or something similar? :)

dev-digital: Fair enough, I would pick _ingest_csv_

NickFitz (approved):
👍

defend_data_capture/supply_chains/management/commands/data_loader.py
dev-digital: @mforner13  This has been confirmed by David Hare. Can you take a look at it, please ?

NickFitz (approved):

mforner13 (approved):
Looks good 👍

PR #54, opened 2021-05-20
Update content and styling

mforner13:
This PR:

- removes references to features not yet ready for release - the data visualisation tool and the supply chain/strategic action summary pages
- updates the font and header in-line with non gov.uk hosted services - removing the GOV UK logo and making the font Arial
- updates the homepage spec to better test the content of the page - the deadline and the description paragraph

Given that the deadline changes dynamically based on the last working day of the current month, I have added a script which calls the util `get_last_working_day_of_a_month`. This is then called within a new custom cypress command.

### Homepage after changes
image

### Task list page after changes
image

### DIFF
--- a/cypress/integration/home_page_spec.js
+++ b/cypress/integration/home_page_spec.js
@@ -27,7 +27,6 @@ describe('The Home Page', () => {
     cy.lastWorkingDay().then(deadline => {
           cy.get('li').contains( `You need to complete your monthly update for 5 supply chains by ${deadline}`)
       })
-    cy.get('li').contains('You need to complete your monthly update for 5 supply chains by')
     cy.get('li').contains(
       'Select a supply chain to provide your regular monthly update or to update wider details.'
     )
--- a/cypress/integration/home_page_spec.js
+++ b/cypress/integration/home_page_spec.js
@@ -20,10 +20,14 @@ describe('The Home Page', () => {
   })
   it('displays the correct text', () => {
     cy.get('h1').contains(
-      `Update supply chain information`
+      'Update supply chain information'
     )
+    cy.get('p').contains("It's important to keep your departmental action plan records up to date. This is so we can work towards constantly improving the UK's supply chain resilience.")
     cy.get('h2').contains('Complete your monthly update')
-    cy.get('li').contains('You need to complete your monthly update for 5 supply chains')
+    cy.lastWorkingDay().then(deadline => {
+          cy.get('li').contains( `You need to complete your monthly update for 5 supply chains by ${deadline}`)
+      })
+    cy.get('li').contains('You need to complete your monthly update for 5 supply chains by')
     cy.get('li').contains(
       'Select a supply chain to provide your regular monthly update or to update wider details.'
     )
--- a/get_last_working_day_this_month.py
+++ b/get_last_working_day_this_month.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python3
+
+from datetime import date
+
+from defend_data_capture.supply_chains.utils import (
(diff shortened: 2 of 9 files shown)


get_last_working_day.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
```
mforner13, line 1: Not sure if this should be in another folder rather than in the root of the project? :)

get_last_working_day_this_month.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
```
mforner13, line 1: Not sure if we'd rather this script be in a separate folder or if it's okay at the root of the project? :)

cypress/integration/home_page_spec.js
```diff
@@ -20,10 +20,14 @@ describe('The Home Page', () => {
...
-      `Update supply chain information`
+      'Update supply chain information'
     )
+    cy.get('p').contains("It's important to keep your departmental action plan records up to date. This is so we can work towards constantly improving the UK's supply chain resilience.")
     cy.get('h2').contains('Complete your monthly update')
-    cy.get('li').contains('You need to complete your monthly update for 5 supply chains')
+    cy.lastWorkingDay().then(deadline => {
+          cy.get('li').contains( `You need to complete your monthly update for 5 supply chains by ${deadline}`)
+      })
+    cy.get('li').contains('You need to complete your monthly update for 5 supply chains by')
```
dev-digital, line 30: We don't want this line anymore, do we ?

get_last_working_day_this_month.py
```diff
@@ -0,0 +1,25 @@
...
+
+from datetime import date
```
dev-digital, line 3: Do we need this ?

get_last_working_day_this_month.py
dev-digital: I would go for _scripts_ folder at the root.
Also we don't need line 1(hash-bang or whatever its called) with .py extension

dev-digital (commented):
Minor comments.

get_last_working_day_this_month.py
NickFitz: Might be worth moving it one level in, alongside `manage.py`? But I daresay it's fine where it is :)

get_last_working_day_this_month.py
NickFitz: Shows as an unused import at my end :)

get_last_working_day_this_month.py
NickFitz: As the Cypress command is running this as `python get_last_working_day_this_month.py` it doesn't need the #! line - that's needed so the shell (e.g. bash) knows how to execute it. So you could remove it, or ensure it's `chmod +x` and have cypress execute it as `cy.exec('get_last_working_day_this_month.py')`. But it's probably easier to leave the Cypress command as-is and not have to bother about marking the file as executable, in which case Python will just treat it as a comment anyway :)

NickFitz (approved):
:+1

get_last_working_day_this_month.py
mforner13: Oops - removed!

cypress/integration/home_page_spec.js
mforner13: Good spot thanks! Removed

get_last_working_day_this_month.py
mforner13: Removed the top line and moved to defend_data_capture/scripts as per our chat in teams :)

dev-digital (approved):

PR #56, opened 2021-05-24
Feature/title tag

NickFitz:
Adds <title>s for each page.

Draws on https://www.gov.uk/service-manual/design/writing-for-user-interfaces#style which has some pointers, and additional discussion at https://github.com/alphagov/govuk-design-system-backlog/issues/157

A couple more fixtures were added, which meant some of the existing Cypress tests had to be tweaked. In order to avoid having to change them again if further fixtures are added in future, I've replaced tests using `contains()` with `invoke('text').matches()` using regular expressions with `\d+` instead of the embedded numerical values that had changed.


NickFitz:
Just realised there's a conflicting change with `main` so changing this to draft status for now.

defend_data_capture/supply_chains/templates/sau_review.html
```diff
@@ -1,6 +1,7 @@
-
 {% extends "base.html" %}
 
+{% block page_title %}Current monthly update - {{ strategic_action.name }} update – {{ supply_chain.name }} – {{ block.super }}{% endblock page_title %}
```
mforner13, line 3: Think you could get rid of the 'update' after the strategic action name since it says it in the first bit :)

cypress/integration/page_titles_spec.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+context('The <title> of', () => {
+
+    describe('The Home Page', () => {
+        it('has the correct title', () => {
+            cy.visit(urls.home);
+            cy.title().should('equal', expectedTitles.home());
+        });
+    });
```
mforner13, line 56: With the tests being called 'has the correct title' I don't think you'd need the context line. Alternatively you could have 'The title of - The {page} - is correct'? :)

cypress/integration/page_titles_spec.js
```diff
@@ -0,0 +1,134 @@
+import supplyChains from '../fixtures/supplyChains.json';
+import strategicActions from '../fixtures/strategicActions.json';
+import strategicActionUpdates from '../fixtures/strategicActionUpdates.json';
+
+const supplyChain = supplyChains.find((supplyChain) => supplyChain.fields.slug === 'supply-chain-7');
+const strategicAction = strategicActions.find((strategicAction) => strategicAction.fields.supply_chain === supplyChain.pk && strategicAction.fields.slug === 'sa-title-test')
+const strategicActionUpdate = strategicActionUpdates.find((strategicActionUpdate) => strategicActionUpdate.fields.strategic_action === strategicAction.pk && strategicActionUpdate.fields.slug === '05-2021');
+
+
+const urls = {
```
mforner13, line 10: It could be really useful to move this out to be accessible by other specs :) Could be in support/utils?

mforner13 (commented):
Just a few suggestions :)

defend_data_capture/supply_chains/templates/sau_review.html
NickFitz: Ah yes, that's left over from the form pages where the update is being created, so not really relevant here :)

cypress/integration/page_titles_spec.js
NickFitz: The context are mainly not necessary, I think, so I've tidied it up a bit :)

cypress/integration/page_titles_spec.js
NickFitz: Yes, good idea :)

NickFitz: Ah, it's a bit more complicated than that… as it needs the various fixtures passed to it, I'll make it callable.

mforner13 (approved):
Looks good 👍

defend_data_capture/supply_chains/templates/sau_review.html
```diff
@@ -1,6 +1,7 @@
-
 {% extends "base.html" %}
 
+{% block page_title %}Current monthly update - {{ strategic_action.name }} – {{ supply_chain.name }} – {{ block.super }}{% endblock page_title %}
```
dev-digital, line 3: Are there any recommendation/restrictions for the length of the title from GDS ?

Would this title show up as ?

```Current monthly update - Strategic action 01 - Ceramics - Update supply chain information```

dev-digital (approved):

defend_data_capture/supply_chains/templates/sau_review.html
NickFitz: No, there's very little advice - just the things I linked in the PR. HMRC recommend no more than four dash-separated components, so although it's a little verbose, it's within that limit.

PR #60, opened 2021-05-25
RT-278: Enable HSTS headers and SSL redirect

NickFitz:
Enable HSTS and SSL redirect, but with an opt-out via an environment variable.


.github/workflows/run-tests.yaml
```diff
@@ -29,5 +29,7 @@ jobs:
         run: docker-compose -f docker-compose.yaml up -d
       - name: Run unit tests
         run: pytest defend_data_capture
+        env:
+          SET_HSTS_HEADERS: ''
```
mforner13, line 33: Would it be a bit clearer if this was `False` rather than an empty string? :)

mforner13 (changes requested):
Looks good - I think it needs an addition to `sample.env` though so new contributors have it as standard when they set up the project.

.github/workflows/run-tests.yaml
NickFitz: Then it will come through as the string "False" which will evaluate to True when Python converts it to a Boolean. Well, I think so; let me check… ah, no, looks like `env` works out it's meant to be a Boolean from the default - change the default to a string and it comes out as the string "False", but with `default=True` it comes out as Boolean `False` :) Will change it…

NickFitz: OK, done and it didn't break :)

mforner13 (approved):
👍

PR #61, opened 2021-05-25
RT-280: Setup test coverage

dev-digital:
**Changes**
- Include _coverage_ related packages to pip requirements.
- Update GH Actions to call coverage and comment the result
- Update gitignore
- Un-related Syntax warning fix for test file.

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -28,8 +28,4 @@ jobs:
       - name: Bring up db container
         run: docker-compose -f docker-compose.yaml up -d
       - name: Run unit tests
-        run: pytest --cache-clear --cov=defend_data_capture defend_data_capture > pytest-coverage.txt
-      - name: Comment coverage
-        uses: coroo/pytest-coverage-commentator@v1.0.2
-      - name: Run functional tests
-        run: bash run_functional_tests.sh
+        run: pytest --cache-clear --cov=defend_data_capture --cov-report=xml defend_data_capture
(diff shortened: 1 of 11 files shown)


.github/workflows/run-tests.yaml
```diff
@@ -28,6 +28,8 @@ jobs:
       - name: Bring up db container
         run: docker-compose -f docker-compose.yaml up -d
       - name: Run unit tests
-        run: pytest defend_data_capture
+        run: pytest --cache-clear --cov=defend_data_capture defend_data_capture > pytest-coverage.txt
+      - name: Comment coverage
+        uses: coroo/pytest-coverage-commentator@v1.0.2
```
dev-digital, line 33: It produces quite a verbose comment to my liking. Happy to disable if appropriate.

NickFitz: I don't think it matters too much if it's verbose - most of the time it won't be seen, but when we _do_ need to look at it, more is probably better :)

NickFitz (approved):
Looks OK to me :)

dev-digital:
> Is there a way to just show the total coverage perhaps? Or just total and then can expand something to see all files? :) Am thinking that having every test listed is really going to take up a lot of space as the app grows.

Exactly, hence my above comment.
Yesterday I found this logged as an [issue](https://github.com/coroo/pytest-coverage-commentator/issues/71) on the tool I am using to produce this comment.

There are ways to publish reports(HTML) from these coverage that I have done in Azure for other projects. While searching for equivalent in GH, I found few actions(one of them is what I have used).

I can explore if GH allows user to publish HTML based reports, if you want

NickFitz (approved):
Nice! 👍

PR #64, opened 2021-05-26
Create healthcheck app with monitoring endpoint

mforner13:
### Purpose of the PR

This PR addressed the need for a healthcheck endpoint as part of the go-live checklist. I have used [the instructions for setting up monitoring](https://readme.trade.gov.uk/docs/howtos/healthcheck.html#healthcheckpattern) in DIT's readme.

Once this PR is merged and we have our production app running, SRE will set up integration with Pingdom. Pingdom will make periodic requests to `our-app/healthcheck` endpoint - which will act as a smoke test for our application.

### How it works

I have created a separate `healthcheck` django app with its own models/migration/views etc, which I thought makes it clear that it's not part of the main functionality of the project.

When a GET request is made to `our-app/healthcheck`:

- The `db_check()` method is called
- Within this method, a query is made to the HealthCheck model
- If the query executes successfully 'OK' will be returned from this method, and along with the response time it will be passed to the template with a 200 HTTP status code.
- If the query fails to execute 'FAIL' will be returned from this method, and along with the response time it will be passed to the template with a 503 HTTP status code.
- The `Cache-Control` and `Content-Type` headers will also be set according to the instructions from the readme.

Screenshot 2021-05-26 at 14 57 23

**To note**
Includes a data migration to ensure a healthcheck object is in the database when first migrations are applied.


defend_data_capture/healthcheck/templates/healthcheck.html
```diff
@@ -0,0 +1,4 @@
+<pingdom_http_custom_check>
+    <status>{{ status }}</status>
+    <response_time>{{ response_time }}</response_time>
+</pingdom_http_custom_check>
```
NickFitz, line 4: I'm very wary of producing XML from a template, but I daresay it'll be OK in this case ;-)

Ref: https://hsivonen.fi/producing-xml/

mforner13: Ooo I didn't know this! I just copied what another django project at DIT had done! What other ways could we produce it? :)

NickFitz: We'd have to either adapt some kind of existing thing that takes a dict or something and turns it into XML, or create an XML DOMDocument object and construct it that way, then serialise the result. Definitely not worth it for this case :)

NickFitz (commented):
Looks good; I'll just wait for the data migration before approving :)

defend_data_capture/healthcheck/checks.py
```diff
@@ -12,7 +12,7 @@ def db_check():
     :return: True or False according to successful retrieval
     """
     try:
-        HealthCheck.objects.get(health_check_field=True)
+        h = HealthCheck.objects.first()
```
NickFitz, line 15: This won't work, I'm afraid - you need to check for `h is None` as the failure condition, or something like that. `first()` won't raise an exception if there's nothing there, it just returns `None`.

mforner13: I think that's fine though. We want this check to fail if the app can't access the database, not if there's no healthcheck's in the db. So if it returns None surely it means it queried the db and there's nothing in there which would be fine :)

NickFitz: Ah, fair enough :)

NickFitz (approved):

defend_data_capture/healthcheck/models.py
```diff
@@ -0,0 +1,11 @@
+from django.db import models
+
+
+class HealthCheck(models.Model):
+    health_check_field = models.BooleanField(null=False, default=True)
+
+    def __unicode__(self):
+        return "%s" % self.health_check_field
```
dev-digital, line 8: Is this necessary with runtime 3.x ?

As per google results, it made sense for 2.x

defend_data_capture/healthcheck/templates/healthcheck.xml
```diff
@@ -0,0 +1,4 @@
+<pingdom_http_custom_check>
```
dev-digital, line 1: Why did we go for xml than html ? more for my knowledge

defend_data_capture/healthcheck/views.py
```diff
@@ -0,0 +1,30 @@
...
+
+from healthcheck.constants import HealthStatus
+
+from healthcheck.checks import db_check
+
+
+class HealthCheckView(TemplateView):
+    template_name = "healthcheck.xml"
+
+    def get_context_data(self, **kwargs):
```
dev-digital, line 13: we could have done this inside _get_, couldn't we ..?
Not that you have to change now.

defend_data_capture/healthcheck/checks.py
```diff
@@ -0,0 +1,19 @@
+import requests
```
dev-digital, line 1: we don't need this, do we ? (and settings)

defend_data_capture/healthcheck/checks.py
```diff
@@ -0,0 +1,19 @@
...
+from sentry_sdk import capture_exception
+
+from healthcheck.constants import HealthStatus
+from healthcheck.models import HealthCheck
+
+
+def db_check():
+    """
+    Performs a basic check on the database by performing a select query on a simple table
+    :return: True or False according to successful retrieval
```
dev-digital, line 12: We returning str, yes ?

defend_data_capture/healthcheck/test/test_healthcheck.py
```diff
@@ -0,0 +1,43 @@
...
+        assert response.context["response_time"] > 0
+        assert response.headers["Content-Type"] == expected_HTTP_headers["Content-Type"]
+        assert (
+            response.headers["Cache-Control"] == expected_HTTP_headers["Cache-Control"]
+        )
+
+    @pytest.mark.django_db()
+    def test_200_response(self, expected_HTTP_headers):
+        client = Client()
+        h = HealthCheck(health_check_field=True)
```
dev-digital, line 34: In real life, we need to load this object, correct ? 
Would that not be part of this PR ?

defend_data_capture/healthcheck/views.py
```diff
@@ -0,0 +1,30 @@
+import time
+
+from django.views.generic import TemplateView
+
+from healthcheck.constants import HealthStatus
+
```
dev-digital, line 6: Lets have [_isort_](https://pypi.org/project/isort/) for such things .. ?

dev-digital (commented):
Minor comments

defend_data_capture/healthcheck/templates/healthcheck.xml
dev-digital: To answer my own question.. :grimacing: 
image

mforner13: The response being in xml was one of the criteria from the go live checklist. I'm guessing it must be what Pingdom expects :) 

Screenshot 2021-05-27 at 14 55 44

defend_data_capture/healthcheck/views.py
mforner13: Ah good spot have corrected :) But yeah could be good to include isort!

defend_data_capture/healthcheck/test/test_healthcheck.py
dev-digital: Soon as I see migration files, I have been clicking the tick until now :smile:

defend_data_capture/healthcheck/views.py
mforner13: I'm not sure if either is better tbh! Having the methods separately separates out the logic, as the `get()` method just handles the response status and headers this way.

dev-digital (approved):

PR #70, opened 2021-05-28
Add cookie settings

mforner13:
This PR adds the following cookie-related settings to `settings.py` on advice of Mateusz (chief architect) and Liam (cyber security), all of which will default to True unless an env var is set.

```
CSRF_COOKIE_SECURE  #Recommended by Liam
CSRF_COOKIE_HTTPONLY  #Recommended by Liam
SESSION_COOKIE_SECURE  #Recommended by Liam
SESSION_COOKIE_AGE  #determines how long the user's session will last (default to 10 hours)
```

Info about what each setting does can be found [here in the django docs](https://docs.djangoproject.com/en/3.2/ref/settings/#session-cookie-age).


dev-digital (approved):
:+1:

PR #73, opened 2021-06-01
RT-294: Include config for dependabot with ignore webpack-dev-server

dev-digital:


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: webpack-dev-server
+      - dependency-name: ws
+    rebase-strategy: disabled
```
munhitsu, line 10: you are missing a new line here

munhitsu (changes requested):

.github/dependabot.yml
dev-digital: @munhitsu Thanks for the input. I see that you added a new line in last commit but alert has not gone away. 



BTW, Even before this commit, PR's check was happy with config validation.

image

PR #74, opened 2021-06-01
Wipe local db before loading fixtures & fixup sample env

mforner13:
This PR wipes the local db using `manage.py flush` before loading fixtures in the `make set-up db` command. Since there is now a data migration which adds a DIT gov department to the db, this addresses the issue of having two duplicate departments in the local db after running `make create-db`.

This PR also includes commenting out the `SESSION_COOKIE_AGE` env var in sample.env, as it doesn't have a value assigned to it.


mforner13:
Tests currently failing due to dates in fixtures not having updated to new month.

Makefile
```diff
@@ -4,6 +4,8 @@ setup:
 setup-db: setup
 	python update_supply_chain_information/manage.py migrate
 	python update_supply_chain_information/manage.py createinitialrevisions
+	# Remove the DIT gov department added in a data migration
+	python update_supply_chain_information/manage.py flush --no-input
```
dev-digital, line 8: This is how I would address data migration side effects on unit tests, if need be :smile:

dev-digital (approved):

Makefile
NickFitz: Yes, this is just the thing for ensuring the DB is clean for things like tests, which is presumably why it's there 😁

NickFitz (approved):
👍

Makefile
```diff
@@ -17,7 +19,7 @@ tests:
 	pytest update_supply_chain_information
 
 load-data: setup
-	python update_supply_chain_information/manage.py loaddata cypress/fixtures/*.json
+	python update_supply_chain_information/manage.py loaddata cypress/fixtures/*.json
```
NickFitz, line 22: There seems to be a stray space at the end of this line :)

NickFitz:
If the tests are failing due to the date thing, we should probably move the data fix over to this patch? Because the date fix doesn't work without this 😆

It's an import statement and a single line, so I could send it across?

mforner13:
@NickFitz moving this over to your work sounds good :) Let me know when you're happy for me to close this PR

NickFitz:
OK, I've merged this branch into my working branch so you can ditch this PR now 🙂

PR #84, opened 2021-06-03
Add scripts for google tag manager

mforner13:
This PR adds the `<script>` and `<noscript>` tags recommended by the performance analysis team to the head and body of the base template. This will allow the app to send events to google tag manager.

As the auth string for google tag manager is different for each environment (and there isn't one for production), I have made it an environment variable called `GTM_AUTH_STRING`.

To allow the base template to access this env var, I've created a custom template filter which will provide the value for a given environment variable.

### DIFF
--- a/update_supply_chain_information/supply_chains/templates/base.html
+++ b/update_supply_chain_information/supply_chains/templates/base.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 {% load render_bundle from webpack_loader %}
 {% load webpack_static from webpack_loader %}
+{% load supply_chain_filters %}
 <html lang="en" class="govuk-template ">
 
 <head>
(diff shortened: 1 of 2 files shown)


update_supply_chain_information/supply_chains/templatetags/supply_chain_filters.py
```diff
@@ -0,0 +1,6 @@
+from django.template.defaulttags import register
+import os
```
dev-digital, line 2: Minor: order of imports

update_supply_chain_information/supply_chains/templates/base.html
```diff
@@ -12,9 +13,22 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <link rel="shortcut icon" sizes="16x16 32x32 48x48" href="{% webpack_static 'favicon.ico' %}" type="image/x-icon"/>
     {% render_bundle 'main' 'css' %}
+    <!-- Google Tag Manager -->
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    'https://www.googletagmanager.com/gtm.js?id='+i+dl+'{{ "GTM_AUTH_STRING"|env }}';f.parentNode.insertBefore(j,f);
+    })(window,document,'script','dataLayer','GTM-M5ZZ7ZS');</script>
```
dev-digital, line 21: Is this js from sample code  or other GDS apps?

dev-digital (approved):
:+1:

update_supply_chain_information/supply_chains/templates/base.html
mforner13: It's what the performance analyst, Dom sent over for us to add :) I think it's pretty standard but links to our apps's container ID (GTM-M5...). I double checked with him and this isn't sensitive as is easily seen in the dev tools on the app apparently.

NickFitz (approved):
👍

PR #85, opened 2021-06-03
RT-305: Fix/enable tests

dev-digital:
**Changes**
- Re-enable _datafixup_ tests
- Ask for confirmation while fixing data  using _datafixup_
- To alleviate testDB related over-rides to Django's internal working, using dev-server/dev-db for running functional tests.
- Reset DB before and after running functional tests
- Disable TTY for _exec_ commands within docker container(to work in GHA)

### DIFF
--- a/run_functional_tests.sh
+++ b/run_functional_tests.sh
@@ -9,15 +9,14 @@ docker-compose -f docker-compose.yaml -f docker-compose.override.yaml up -d
 # Save the pid for the server to a file 'backend.pid'
 # Run all cypress tests after a 5 second delay to allow the django server to be brought up
 export SET_HSTS_HEADERS='False'
-python update_supply_chain_information/manage.py testserver \
-    --addrport 8001 \
+
+# reset db before test run
+make drop-db && make create-db
+
+python update_supply_chain_information/manage.py runserver \
     --settings=test_settings \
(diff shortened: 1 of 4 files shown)


run_functional_tests.sh
```diff
@@ -9,10 +9,9 @@ docker-compose -f docker-compose.yaml -f docker-compose.override.yaml up -d
 # Save the pid for the server to a file 'backend.pid'
 # Run all cypress tests after a 5 second delay to allow the django server to be brought up
 export SET_HSTS_HEADERS='False'
+export DATABASE_URL=postgres://postgres:password@localhost:5432/test_supply_chain_info
```
dev-digital, line 12: Nice and clean, I like it :+1:

mforner13: Ah no that's a good point! We can just put the `DATABASE_URL` in the test settings, as the make command simply creates the db! Will update :)

mforner13: Tested it out, and the database_url does indeed need to be set before running `make create-db` so that the fixtures are fixed in the right database :)

mforner13 (approved):
👍

NickFitz (approved):
Looks good 👍

PR #101, opened 2021-06-11
Add receive_feedback_emails field to User model

mforner13:
This PR adds a new field `receive_feedback_emails` to the User model, which defaults to False. 

When set to True for a user (will just be for DIT admin users), they will receive feedback emails from users. This prevents the need to hard code any email addresses.


NickFitz (approved):
Looks OK to me 👍

dev-digital (approved):

PR #102, opened 2021-06-11
RT-318: Links in monthly update form navigation

NickFitz:
This implements the basic version of the modified top navigation, where there's no link to pages after the current page.

<s>I'll start looking into the matter of linking to pages beyond the current page if those later pages have already been completed, but this should be adequate for an initial release.</s>

It  now also implements  linking to pages beyond the current page if those later pages have already been completed.

<s>I'll leave this PR in draft for now.</s>


update_supply_chain_information/supply_chains/templates/supply_chains/includes/monthly_update_form_breadcrumbs.html
```diff
@@ -2,9 +2,15 @@
...
             <li class="moj-sub-navigation__item">
-                <a class="moj-sub-navigation__link"{% if link_data.is_current_page %} aria-current="page"{% endif %} href="{{ link_data.url }}">
-                    {{ forloop.counter }}. {{ link_data.label }}
-                </a>
+                {% if link_data.not_a_link %}
+                    <span class="moj-sub-navigation__link"{% if link_data.is_current_page %} aria-current="page"{% endif %}>
+                        {{ forloop.counter }}. {{ link_data.label }}
+                    </span>
+                {% else %}
+                    <a class="moj-sub-navigation__link"{% if link_data.is_current_page %} aria-current="page"{% endif %} href="{{ link_data.url }}">
```
mforner13, line 10: And the opposite for the items which are links - would we want them to become blue when hovering? Currently they're in grey so could be unclear that you can click on them. Although I can't work out why the blue on hover is being applied to the non-links and not these when they seem to have the same classes!

update_supply_chain_information/supply_chains/models.py
```diff
@@ -371,6 +368,62 @@ def is_changing_target_completion_date(self):
     def has_no_timing_information(self):
         return self.has_no_target_completion_date and self.has_no_is_ongoing
 
+    @property
+    def has_initial_timing(self):
```
mforner13, line 372: To me 'initial' implies it was already there - would `has_new_timing` be clearer? :)

update_supply_chain_information/supply_chains/models.py
```diff
@@ -371,6 +368,62 @@ def is_changing_target_completion_date(self):
...
+        return self.has_new_target_completion_date or self.has_new_is_ongoing
+
+    @property
+    def has_existing_timing(self):
+        return self.has_existing_target_completion_date or self.is_currently_ongoing
+
+    @property
+    def has_revised_timing(self):
+        return (
+            self.has_existing_target_completion_date or self.is_currently_ongoing
```
mforner13, line 382: Could you replace this line with `self.has_existing_timing` ? :)

update_supply_chain_information/supply_chains/test/test_models.py
```diff
@@ -645,3 +646,408 @@ def test_user_changing_target_completion_date_is_recorded_by_reversion(
...
         )
         assert version.revision.user.email == expected_user_email
+
+    def test_has_action_status_true_if_has_implementation_rag_rating_green(self):
+        self.strategic_action_update.implementation_rag_rating = RAGRating.GREEN
+        self.strategic_action_update.save()
+        self.strategic_action_update.refresh_from_db()
+        assert self.strategic_action_update.action_status_complete
+
+    def test_has_action_status_true_if_has_implementation_rag_rating_amber_and_reason_for_delays(
```
mforner13, line 656: Could any of these new tests be parametrised? Perhaps a parametrised test per model property e.g. one for `has_action_status`, one for `has_revised_timing` etc... Could really help to reduce repetition and keep the file more of a manageable size :)

update_supply_chain_information/supply_chains/test/test_views.py
```diff
@@ -211,7 +211,7 @@ def test_posting_form_saves_changed_target_completion_date_to_strategic_action_u
             )
 
 
-class TestNoCompletionDateMonthlyUpdateNavigationLinks:
+class TestNoCompletionDateMonthlyUpdateNavigationItems:
```
dev-digital, line 214: We could better organise these tests as update related tests has out grown, either 

-  Create a new test file for Monthly updates and move tests to that, or
- Create GH ticket

cypress/integration/monthly_update_forms_with_completion_date_spec.js
```diff
@@ -244,16 +244,38 @@ describe('Testing monthly update forms', () => {
                 })
               })
             })
-            context('The link marked as the current page should be', () => {
+            context('The breadcrumb marked as the current page should be', () => {
```
dev-digital, line 247: Hmm, I see that few more lines added to 900 lines long test scenario and it will only _grow_
(with imminent change on expected completion date .. ?)

How do we handle this ?

dev-digital (commented):
Added few comments

update_supply_chain_information/supply_chains/templates/supply_chains/includes/monthly_update_form_breadcrumbs.html
```diff
@@ -2,9 +2,15 @@
     <ol class="moj-sub-navigation__list">
         {% for _, link_data in navigation_links.items %}
             <li class="moj-sub-navigation__item">
-                <a class="moj-sub-navigation__link"{% if link_data.is_current_page %} aria-current="page"{% endif %} href="{{ link_data.url }}">
-                    {{ forloop.counter }}. {{ link_data.label }}
-                </a>
+                {% if link_data.not_a_link %}
+                    <span class="moj-sub-navigation__link"{% if link_data.is_current_page %} aria-current="page"{% endif %}>
```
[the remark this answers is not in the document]
NickFitz: I had a fix for that before, but it must have got lost in the mix. Will work it out again and re-apply it :)

update_supply_chain_information/supply_chains/templates/supply_chains/includes/monthly_update_form_breadcrumbs.html
NickFitz: Ah, probably got the fix the wrong way round then :D

update_supply_chain_information/supply_chains/models.py
NickFitz: Yes, sounds good :)

update_supply_chain_information/supply_chains/models.py
NickFitz: I could :D

update_supply_chain_information/supply_chains/test/test_models.py
NickFitz: I thought of that, but it would involve me learning all the ins and outs of parametrising tests, and after spending some time working out what would be involved, I decided getting the fix ready for go-live was a better use of the limited time available :)

update_supply_chain_information/supply_chains/test/test_models.py
```diff
@@ -645,3 +646,408 @@ def test_user_changing_target_completion_date_is_recorded_by_reversion(
             .last()
         )
         assert version.revision.user.email == expected_user_email
+
+    def test_has_action_status_true_if_has_implementation_rag_rating_green(self):
+        self.strategic_action_update.implementation_rag_rating = RAGRating.GREEN
+        self.strategic_action_update.save()
+        self.strategic_action_update.refresh_from_db()
```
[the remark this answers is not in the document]
NickFitz: Potentially, as any additional actions taken in the `save()` method that modify the object's state wouldn't be reflected in the existing instance. (For example, if I got an instance, set its `slug` field to the empty string, saved it, then examined the slug field of the instance, it would still be the empty string even though the `save()` method would have re-created the value for `slug` and saved it in the database.) I generally feel it's worth making sure the object one has is definitely the same as the object in the database, particularly in tests.

update_supply_chain_information/supply_chains/test/test_monthly_update_forms_navigation.py
```diff
@@ -0,0 +1,984 @@
...
+        target_completion_date=date.today() + relativedelta(months=2),
+        is_ongoing=False,
+    )
+    return strategic_action
+
+
+@pytest.fixture
+def strategic_action_with_is_ongoing():
+    supply_chain = SupplyChainFactory()
+    strategic_action: StrategicAction = StrategicActionFactory(
```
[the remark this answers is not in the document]
NickFitz: Usually, the type is there if I wanted to get autocomplete working for that variable on a subsequent line ;)

update_supply_chain_information/supply_chains/test/test_monthly_update_forms_navigation.py
```diff
@@ -0,0 +1,984 @@
...
+    def test_update_info_view_not_linked_to_self_when_not_completed(
+        self,
+        strategic_action_update_info_timing_status_incomplete,
+        logged_in_client,
+        test_user,
+    ):
+        test_user.gov_department = (
+            strategic_action_update_info_timing_status_incomplete.supply_chain.gov_department
+        )
+        test_user.save()
```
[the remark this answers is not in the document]
NickFitz: Moved up into the three strategic action fixtures 👍

update_supply_chain_information/supply_chains/test/test_views.py
NickFitz: I think it would be a good idea to re-organise the tests once we have a bit less urgency about current work 👍

cypress/integration/monthly_update_forms_with_completion_date_spec.js
NickFitz: It's tricky with these forms because what happens on one page affects other pages. But when there's time, it would be worth identifying the possible scenarios and breaking the tests up into individual suites for each of them.

update_supply_chain_information/supply_chains/test/test_models.py
mforner13: Fair enough! Perhaps parametrising them could be a task for post-go live? :)

update_supply_chain_information/supply_chains/views.py
```diff
@@ -262,6 +262,7 @@ class MonthlyUpdateMixin:
     model = StrategicActionUpdate
     context_object_name = "strategic_action_update"
     slug_url_kwarg = "update_slug"
+    object: StrategicActionUpdate = None
```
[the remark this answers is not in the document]
NickFitz: In the `get_object()` method of the views that use it - they all have the same model, as defined at line 262.

update_supply_chain_information/supply_chains/templates/supply_chains/includes/monthly_update_form_breadcrumbs.html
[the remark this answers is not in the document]
NickFitz: Well, turns out I do indeed have a fix, but I hadn't set webpack running so couldn't see it, as it wasn't in the static files compiled on a different branch - you probably need to do the same :)

update_supply_chain_information/supply_chains/templates/supply_chains/includes/monthly_update_form_breadcrumbs.html
NickFitz: See above :D

update_supply_chain_information/supply_chains/test/test_models.py
NickFitz: Yes, Prashanth talks further down about some other test refactoring that could be done. I think now we've got a better idea of what needs testing, we should spend a little time getting a coherent structure for all the tests sorted out :)

update_supply_chain_information/supply_chains/templates/supply_chains/includes/monthly_update_form_breadcrumbs.html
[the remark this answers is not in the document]
mforner13: Ahh ofc! Pesky webpack :p Looks good!

update_supply_chain_information/supply_chains/models.py
```diff
@@ -375,6 +372,60 @@ def is_changing_target_completion_date(self):
...
+    def has_new_timing(self):
+        return self.has_new_target_completion_date or self.has_new_is_ongoing
+
+    @property
+    def has_existing_timing(self):
+        return self.has_existing_target_completion_date or self.is_currently_ongoing
+
+    @property
+    def has_revised_timing(self):
+        return (self.has_existing_timing) and (
```
mforner13, line 385: I think can remove the brackets around `self.has_existing_timing` ? :)

mforner13 (approved):
Looks good 👍

cypress/integration/monthly_update_forms_with_completion_date_spec.js
dev-digital: Yes, can we create tickets within GH for these(refactoring and re-structring,  ) as part of this PR.

I think, refactoring needs addressing pretty soon reason being, monthly update forms the bulk of current functionality and even without further enhancement there will be maintenance work.

dev-digital (approved):
:+1:

update_supply_chain_information/supply_chains/models.py
NickFitz: Yes, I think they were left over from when I was wrestling with getting the right conditions in place ;)

PR #104, opened 2021-06-14
Add custom access denied page

mforner13:
This PR creates a custom access denied page, returned whenever the `PermissionDenied` exception is raised. 

Django automatically renders a template with the name `403.html` (if it exists) when `PermissionDenied` is raised.

This page will be shown if a user tries to access a URL not related to an object linked to their gov department - e.g. a supply chain their department isn't responsible for.

In order to feed the feedback email addresses to the `mailto:` link in the 403 template, I have added a template tag which will return the email addresses of all users with `receive_feedback_emails` as True, as a comma separated string.

**Access denied page**
Screenshot 2021-06-14 at 11 16 47

### DIFF
--- a/cypress/integration/access_denied_spec.js
+++ b/cypress/integration/access_denied_spec.js
@@ -1,0 +1,29 @@
+import supplyChains from '../fixtures/supplyChains.json'
+import users from '../fixtures/user.json'
+
+const supplyChain = supplyChains[7].fields
+const adminUser = users[1].fields
+
+describe('The access denied page', () => {
+  it('successfully loads', () => {
+    cy.visit(Cypress.config('baseUrl') + `/${supplyChain.slug}/`, {failOnStatusCode: false})
+    cy.injectAxe()
+  })
+  it('has no accessibility issues', () => {
+    cy.runA11y()
+  })
+  it('Displays the correct text', () => {
+      cy.get('h1').contains('Access denied')
+      cy.get('p').contains('You do not have permission to access this page.')
+      cy.get('p').contains('Please try one of the following:')
+      cy.get('ul').children().should('have.length', 4)
+      cy.get('li').contains('If you typed the web address, check it is correct')
+      cy.get('li').contains('If you pasted the web address, check you copied the entire address')
+      cy.get('li').contains('Browse the homepage to find the information you need')
+      cy.get('li').contains('Contact the Resilience Tool team if you think you do have permission to access this page.')
+  })
+  it('Displays correct links', () => {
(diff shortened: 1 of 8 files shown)


cypress/integration/access_denied_spec.js
```diff
@@ -0,0 +1,29 @@
+import supplyChains from '../fixtures/supplyChains.json'
+import users from '../fixtures/user.json'
+
+const supplyChain = supplyChains[7].fields
+const adminUser = users[1].fields
+
+describe('The Home Page', () => {
```
NickFitz, line 7: Should this say something different? ;-)

cypress/integration/access_denied_spec.js
```diff
@@ -0,0 +1,29 @@
...
+  it('successfully loads', () => {
+    cy.visit(Cypress.config('baseUrl') + `/${supplyChain.slug}`, {failOnStatusCode: false})
```
NickFitz, line 9: There should be a / at the end of the URL (see `urls.py` line 98).

cypress/integration/access_denied_spec.js
mforner13: Haha yes it should....copy and pasting shame!

NickFitz (approved):
👍 +1

update_supply_chain_information/supply_chains/templates/403.html
```diff
@@ -0,0 +1,22 @@
...
+<p class="govuk-body">You do not have permission to access this page.</p>
+<p class="govuk-body">
+    Please try one of the following:
+    <ul class="govuk-list govuk-list--bullet">
+        <li>If you typed the web address, check it is correct</li>
+        <li>If you pasted the web address, check you copied the entire address</li>
+        <li>
+            Browse the <a class="govuk-link" href="{% url 'index' %}">homepage</a>
+             to find the information you need</li>
+        {% get_feedback_emails_as_string as feedback_emails %}
```
dev-digital, line 14: More of the question:
Not sure if we need to load custom filters at the top... of if this line is doing that magic ?

dev-digital:
So once any/all admin users login, db will retain records through migrations and further deployments, I take it.

No action required as of now.
Not a fan of data migrations, but I thought there will be some within in PR which will make certain that admin users exist within the system.

dev-digital (approved):
:+1:

update_supply_chain_information/supply_chains/templates/403.html
mforner13: It seems to be able to access the tag without loading it. However I've added the load line in base.html as it seems in the docs on custom tags that it is indeed needed!

PR #105, opened 2021-06-14
Add feedback link & fix alignment in BETA banner

mforner13:
This PR makes two changes to the BETA phase banner in the base template:

1. Adds a mailto link to 'feedback' which will open an email to a group email address with users with `receive_email_feedback` as True bcc'd.
2. Vertically centres the text in the banner (caused by an unneeded <span>)

It updates the feedback link in the 403 template to create the same email (group email in 'to' field, with admin users bcc'd)

It also removes the custom class `app-phase-banner-content` which is no longer needed (left over from when there was an additional link in the right of the banner).

**Updated banner**

Screenshot 2021-06-14 at 15 27 25

### DIFF
--- a/cypress/integration/access_denied_spec.js
+++ b/cypress/integration/access_denied_spec.js
@@ -24,6 +24,6 @@ describe('The access denied page', () => {
   })
   it('Displays correct links', () => {
       cy.get('a').contains('homepage').should('have.attr', 'href').and('eq', '/')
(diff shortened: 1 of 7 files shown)


dev-digital (approved):

NickFitz (approved):
👍 +1

dev-digital (approved):

PR #109, opened 2021-06-15
Fix homepage content

mforner13:
This PR makes two fixes to content on the homepage:

1. It pluralises 'supply chain' only when there are more than 1 incomplete supply chains for that department
2. It moves the sentence 'All your supply chains are complete...' to a paragraph rather than a bullet point

**When all complete**

Screenshot 2021-06-15 at 17 31 12

**When just 1 supply chain**

Screenshot 2021-06-15 at 17 31 50

**When more than 1 supply chain**

Screenshot 2021-06-15 at 17 32 13


NickFitz (approved):
👍

dev-digital (approved):

PR #113, opened 2021-06-17
Make supporting_organisations array field blank-able

mforner13:
Previously, `blank=True` was only applied to the `Charfield` within the Arrayfield. This meant that the supporting_organisations field was mandatory.

This PR applies `blank=True` to the arrayfield as some strategic actions won't have any other organisations supporting their work. Blank=True is still applied to the Charfield within the Arrayfield to allow for empty text inputs to be submitted when the field is rendered in a form.


NickFitz (approved):
Handy to know that obscure little gotcha for future reference :)

dev-digital (approved):

PR #117, opened 2021-06-17
Filter out archived supply chains/strategic actions from views

mforner13:
This PR makes two fixes:

1. Filters out archived supply chains from the home page view
2. Filters out archived strategic actions from the task list view

I have also added a test for each of these scenarios to the relevant files.

In the `SCTaskListView` I have also filtered out archived strategic actions when calculating `self.ready_to_submit_updates` and `self.submitted_only_updates`. 

This means that the numbers displayed in `x out of x strategic actions are not ready to submit` in the template would remain correct even if an update was given for an action since the last deadline, but that action was then archived (and therefore would no longer appear in the task list)

### DIFF
--- a/update_supply_chain_information/supply_chains/views.py
+++ b/update_supply_chain_information/supply_chains/views.py
@@ -153,25 +153,35 @@ class SCTaskListView(
             slug=supply_chain_slug, is_archived=False
         )
 
-        sa_qset = StrategicAction.objects.filter(supply_chain=self.supply_chain)
+        sa_qset = StrategicAction.objects.filter(
+            supply_chain=self.supply_chain, is_archived=False
+        )
         self.total_sa = sa_qset.count()
 
         self.sa_updates = self._get_sa_update_list(sa_qset)
 
-        self.ready_to_submit_updates = StrategicActionUpdate.objects.since(
-            self.last_deadline,
-            supply_chain=self.supply_chain,
-            status__in=[
-                StrategicActionUpdate.Status.READY_TO_SUBMIT,
-                StrategicActionUpdate.Status.SUBMITTED,
-            ],
-        ).count()
+        self.ready_to_submit_updates = (
+            StrategicActionUpdate.objects.since(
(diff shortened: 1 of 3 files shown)


update_supply_chain_information/supply_chains/views.py
```diff
@@ -150,25 +153,35 @@ def _extract_view_data(self, *args, **kwargs):
...
-        self.submitted_only_updates = StrategicActionUpdate.objects.since(
-            self.last_deadline,
-            supply_chain=self.supply_chain,
-            status=StrategicActionUpdate.Status.SUBMITTED,
-        ).count()
+        self.submitted_only_updates = (
+            StrategicActionUpdate.objects.since(
+                self.last_deadline,
+                supply_chain=self.supply_chain,
+                status=StrategicActionUpdate.Status.SUBMITTED,
```
NickFitz, line 180: Could the `is_archived=False` be moved into here? Though maybe it's clearer with the separate filter.

update_supply_chain_information/supply_chains/views.py
```diff
@@ -182,6 +195,11 @@ def _extract_view_data(self, *args, **kwargs):
         if self.update_submitted:
             self._update_review_routes()
 
+        print("incomplete updates", self.incomplete_updates)
```
NickFitz, line 198: Should these `print`s still be in here?

NickFitz (approved):
👍

update_supply_chain_information/supply_chains/views.py
mforner13: Well that's embarrassing 😋  Removed!

update_supply_chain_information/supply_chains/views.py
mforner13: I did think about moving it into the since() query, and then thought that filtering out archived was slightly separate functionality so went for more explicit :)

dev-digital (approved):

PR #120, opened 2021-06-18
RT-342: Implement changes to current/changed completion date display in monthly update forms

NickFitz:
…g or revised timing pages.

### DIFF
--- a/cypress/integration/monthly_update_forms_no_completion_date_spec.js
+++ b/cypress/integration/monthly_update_forms_no_completion_date_spec.js
@@ -634,8 +634,9 @@ describe('Testing monthly update forms', () => {
             it('has the correct page header', function() {
               cy.monthlyUpdatePageHeader(this.strategicAction.name).should('exist')
             })
-            it ('warns that there is no expected completion date', () => {
-              cy.get('body').get('h1 ~ .govuk-warning-text').contains("There's no expected completion date for this action.").should('exist')
+            it ('shows the "Adjusted" completion date just specified on the Timing page', function() {
+              cy.get('h1 ~ .govuk-inset-text > h2').contains("Adjusted estimated date of completion").should('exist')
+              cy.get('h1 ~ .govuk-inset-text > h2 + p').contains(targetCompletionDateRepresentation).should('exist')
             })
             it ('shows instructions', () => {
               cy.mainForm().find('legend + .govuk-body > p:first-of-type').contains('When considering if delivery of the strategic action is on track, consider:');
@@ -759,6 +760,18 @@ describe('Testing monthly update forms', () => {
               })
             })
           })
+          context('Returning to the "Timing" page and selecting Ongoing then coming back to "Delivery Status"', function() {
+            beforeEach(function() {
+              cy.visit(this.timingURL)
+              cy.mainForm().fieldLabelled('No').click()
+              cy.mainForm().fieldLabelled('Ongoing').click()
+              cy.mainForm().submitButton().click()
+            })
+            it ('shows the "Adjusted" timing as "Ongoing"', function() {
+              cy.get('h1 ~ .govuk-inset-text > h2').contains("Adjusted estimated date of completion").should('exist')
+              cy.get('h1 ~ .govuk-inset-text > h2 + p').contains('Ongoing').should('exist')
+            })
+          })
         })
         context('The Check Your Answers page', () => {
           beforeEach(() => {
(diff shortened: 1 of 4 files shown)


update_supply_chain_information/supply_chains/models.py
```diff
@@ -209,6 +209,9 @@ def save(self, *args, **kwargs):
     def last_submitted_update(self):
         return self.monthly_updates.last_month()
 
+    def value_for_timing(self):
+        return self.target_completion_date or self.is_ongoing or False
```
dev-digital, line 213: Why do we need _False_ at the end ?

update_supply_chain_information/supply_chains/models.py
```diff
@@ -209,6 +209,9 @@ def save(self, *args, **kwargs):
...
+
```
dev-digital, line 214: Though not directly related to this PR, Somehow we ended with 2 string representations L215 and 218. Can we just have one ?

update_supply_chain_information/supply_chains/models.py
```diff
@@ -429,6 +432,13 @@ def complete(self):
             and self.revised_timing_complete
         )
 
+    def value_for_timing(self):
+        return (
+            self.changed_value_for_target_completion_date
+            or self.changed_value_for_is_ongoing
+            or False
```
dev-digital, line 439: What happens w/o this line ?

update_supply_chain_information/supply_chains/models.py
```diff
@@ -209,6 +209,9 @@ def save(self, *args, **kwargs):
     def last_submitted_update(self):
         return self.monthly_updates.last_month()
 
+    def value_for_timing(self):
```
dev-digital, line 212: By the usage, it appears like returning a boolean is good enough ?
In which case it could be renamed to something like "has_timing_info". Applies to SAU model too.

cypress/integration/monthly_update_forms_no_completion_date_spec.js
```diff
@@ -759,6 +760,18 @@ describe('Testing monthly update forms', () => {
               })
             })
           })
+          context('Returning to the "Timing" page and selecting Ongoing then coming back to "Delivery Status"', function() {
+            beforeEach(function() {
+              cy.visit(this.timingURL)
+              cy.mainForm().fieldLabelled('No').click()
+              cy.mainForm().fieldLabelled('Ongoing').click()
+              cy.mainForm().submitButton().click()
```
dev-digital, line 768: Thinking out loud here.
When we come to refactoring these tests, would it help to use objects with desired value than _click_ action which would change the state of data/db and introduces dependency

On that topic, can you add those issues to GH please ?

dev-digital (commented):
Minor comments

update_supply_chain_information/supply_chains/models.py
NickFitz: I did think of getting rid of that, as it isn't really necessary given `is_ongoing` is a Boolean. I'll remove it.

update_supply_chain_information/supply_chains/models.py
NickFitz: Yes, that happened when Miriam added in the ones suitable for the admin users without removing the ones I'd originally put in that showed things in a way suitable for my own use in the admin. I'll get rid of my old ones.

update_supply_chain_information/supply_chains/models.py
NickFitz: Yes, originally I envisaged using the returned value in the templates, but that turned out to be impractical, so a name change to reflect the fact that the returned value is only evaluated as a Boolean makes sense.

cypress/integration/monthly_update_forms_no_completion_date_spec.js
NickFitz: I'm not sure what you mean. The sole purpose of integration/functional tests is to demonstrate that the app's functionality is operating correctly as a whole, as distinct from unit tests which demonstrate that individual components operate correctly in isolation. If these tests were somehow prevented from making changes to the underlying data, they wouldn't be testing that the app was actually functional. They might prove that the browser was behaving as expected, but that wouldn't say anything about the correct functioning of the app.

dev-digital: Resolving this comment with intended work to be taken within #123

dev-digital (approved):
:+1:

PR #121, opened 2021-06-21
RT-337: Handling archived objects

dev-digital:
**Description**
Continuation of #117 

**Changes**
- Include archive flag while counting number of SA's within HomePageView
- Test to catch related errors
- Refactor env var _FEEDBACK_GROUP_EMAIL_ to respective modules away from _run_funcationl_test.sh_ so that tests can be run in different configuration


NickFitz (approved):
👍

PR #128, opened 2021-06-24
RT-350: Exclude archived SC within SCCompleteView

dev-digital:
Yet another occurrence of incorrect summary message when archived SC is involved.

Applying the same fix to add a filter to exclude archived SC.

**Note**
As its a minor bug w/o any side affects, target branch is _develop_ intended for next release.


NickFitz (approved):
👍

PR #132, opened 2021-06-28
RT-352: Move to next supported version of node to keep up with CF release 1.7.55

dev-digital:
Move to node 14.17.1 as the current runtime is dropped from the CF supported versions list 
https://github.com/cloudfoundry/nodejs-buildpack/issues/330


NickFitz (approved):
👍

Issue #147, opened 2021-07-09
Refactor supply_chains app to move common code outside of the app 


dev-digital:
**Is your feature request related to a problem? Please describe.**
One of the apps within form application, _supply_chains_ has many common code like

- base template
- privacy statements
- access error templates etc

**Describe the solution you'd like**
Move these common code away from the supply_chains to be used across different apps, say visualisation

dev-digital:
Considering the amount of (potential) common code to be used across visualisation and Forms app and possibility of 

- Amount of common code
- Potential for further changes on them(anticipated to be less)

it was agreed that maintaining copy of the common code would be effective solution at this point in time. Related work is captured within #163

PR #163, opened 2021-07-20
RT-389: Creating common_base template to be used across resilience tools

dev-digital:
**Description**

Work related to RT-389 and #147 

**Changes**
- Create a new template _common_base_ to be used across resilience tools(Forms and Visualisations)

### DIFF
--- a/update_supply_chain_information/supply_chains/templates/common_base.html
+++ b/update_supply_chain_information/supply_chains/templates/common_base.html
@@ -1,8 +1,9 @@
+<!DOCTYPE html>
 {% comment %}
 This template will contain common base components to be used across Forms and Visualisation apps. A copy of this will be maintained within visualisation repo(RT-389) and expected to be in sync with this file. Any change will have to be applied to both places.
 {% endcomment %}
 
-<!DOCTYPE html>
+
 {% load render_bundle from webpack_loader %}
 {% load webpack_static from webpack_loader %}
 <html lang="en" class="govuk-template ">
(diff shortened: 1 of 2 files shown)


update_supply_chain_information/supply_chains/templates/common_base.html
```diff
@@ -0,0 +1,108 @@
+{% comment %}
+This template will contain common base components to be used across Forms and Visualisation apps. A copy of this will be maintained within visualisation repo(RT-389) and expected to be in sync with this file. Any change will have to be applied to both places.
+{% endcomment %}
+
+<!DOCTYPE html>
```
NickFitz, line 5: I was wondering if it would be better to move the doctype above the comment above, though the current version of the HTML spec says whitespace before it is permitted. I know having the doctype start anywhere but the first character in the document (not counting a BOM) used to cause problems on Internet Explorer (though I'm not sure we need to worry about that now). The GDS page template at https://design-system.service.gov.uk/styles/page-template/ has it starting at the first character (no whitespace before it), and https://design-system.service.gov.uk/get-started/updating-your-code/ suggests doing that.

NickFitz (changes requested):
Just the one thing :)

update_supply_chain_information/supply_chains/templates/common_base.html
dev-digital: Done

NickFitz (approved):
👍

PR #170, opened 2021-07-23
RT-413: Remove the `SECURE_SSL_REDIRECT` setting (which isn't useful anyway)

NickFitz:
The `SECURE_SSL_REDIRECT` setting ensures all non-HTTPS requests are 301 redirected to HTTPS. However, this redirect is already done by PaaS for all clients attempting to access the app via externally-visible domains.

Activity Stream (and potentially other internal apps in the future) want to be able access the app via internal networking using HTTP, going directly to the Gunicorn app on port 8080. The domain `update-supply-chain-information.apps.internal` has been configured in PaaS and in `ALLOWED_HOSTS` in Vault to allow this access. But the `SECURE_SSL_REDIRECT` setting causes the Django app to redirect Activity Stream to HTTPS, and it isn't configured to handle this redirect.

As the necessary redirect is already issued to external clients by PaaS, and the setting for the Django app therefore isn't being used anyway except for internal clients which don't want or need it, it has been disabled.


dev-digital:
> it isn't configured to handle this redirect

To which URL/Domain does the activity client ping the request to _update-supply-chain-information.london.cloudapps.digital_ ?

dev-digital (approved):

PR #174, opened 2021-07-26
RT-415: Ensure the user is added to a Strategic Action Update when they start one

NickFitz:
SAUs weren't saving the user that created them in their `user` field.


dev-digital (approved):
:+1:

PR #187, opened 2021-08-12
RT-434: Fix/activity stream data loss

NickFitz:
Some models have many instances with identical values for `last_modified`. The `activity_stream` pagination depends on ordering by that field. As the instances are unordered within the ordering by `last_modified`, it becomes possible for any given instance to appear on multiple pages, or not to appear on any page.

This can be addressed by adding `id` as a second-level ordering.

### DIFF
--- a/update_supply_chain_information/activity_stream/test/test_pagination.py
+++ b/update_supply_chain_information/activity_stream/test/test_pagination.py
@@ -203,3 +203,44 @@ class TestActivityStreamCursorPagination:
             second_page_items = pagination.paginate_queryset(queryset, drf_request)
             for item in second_page_items:
                 assert item["json"]["id"] not in first_page_item_ids
+
+    @mock.patch(
+        "django.utils.timezone.now",
+        mock.MagicMock(
+            return_value=datetime.datetime(
+                year=2021, month=6, day=21, hour=3, minute=32, second=17
+            )
+        ),
+    )
+    def test_paginated_response_includes_all_items(self, supply_chain, rf):
+        """
+        When the data spans more than one page, those models which have more than `page_size` instances
+        for which `last_modified` has the same value
+        are not having all instances included, as they are replaced by duplicate occurrences of other instances.
+        This was due to the pagination only ordering by `last_modified`.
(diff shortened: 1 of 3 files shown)


update_supply_chain_information/activity_stream/test/test_pagination.py
```diff
@@ -154,3 +156,91 @@ def test_paginated_response_has_no_next_link_on_empty_page(
...
+                page_length * 2, supply_chain=supply_chain
+            )
+            all_last_modified = [
+                strategic_action.last_modified
+                for strategic_action in StrategicAction.objects.all()
+            ]
+            all_last_modified_set = {
+                strategic_action.last_modified
+                for strategic_action in StrategicAction.objects.all()
+            }
```
dev-digital, line 191:
```
set(all_last_modified)
```
is more explicit

update_supply_chain_information/activity_stream/test/test_pagination.py
```diff
@@ -154,3 +156,91 @@ def test_paginated_response_has_no_next_link_on_empty_page(
...
+            page_item_ids = {item["json"]["id"] for item in first_page_items}
+            next_link = pagination.get_next_link()
+            while next_link:
+                request = rf.get(next_link)
+                drf_request = Request(request)
+                next_page_items = pagination.paginate_queryset(queryset, drf_request)
+                next_page_item_ids = {item["json"]["id"] for item in next_page_items}
+                page_item_ids = page_item_ids | next_page_item_ids
+                next_link = pagination.get_next_link()
+            assert all_item_ids.issubset(page_item_ids)
```
dev-digital, line 246: Minor:

You are better off going for equality here as all ids are expected to be present in *page_item_ids* if I read the script correctly.

dev-digital (approved):
Minor comments.

I presume you want to deploy this to master which would be another PR ?

update_supply_chain_information/activity_stream/test/test_pagination.py
NickFitz: Not quite: one or more of the pages will also include the IDs of the default GovDepartment and of the SupplyChain fixture (and I think a User too). So the `page_item_ids` set will end up with a small number of IDs that aren't in the StrategicAction IDs set.

PR #190, opened 2021-08-13
Release changes for RT-434 to production

NickFitz:


dev-digital (approved):
@NickFitz 
So in your release notes/details you would list this and user bug fix(and any other changes that has gone into main since last release), correct ?

NickFitz:
Yes, will do 👍

PR #197, opened 2021-08-19
RT-435: Feature/sap filters

dev-digital:
**Changes**
- New app *action_progress* added to implement SAP details
- Created Base and couple of derived classes to implement required view with 2 nested filters and a Form
- Tests
- Inclusion of custom *404* page

**Screenshots**

image

### DIFF
--- a/cypress/integration/sap_filters_spec.js
+++ b/cypress/integration/sap_filters_spec.js
@@ -1,0 +1,195 @@
+import users from '../fixtures/user.json'
+import govDepartments from '../fixtures/govDepartment.json'
+import supplyChains from '../fixtures/supplyChains.json'
+import { urlBuilder } from "../support/utils.js"
+
+const adminUser = users[0].fields
+const govDepartment = govDepartments[0].fields
+const supplyChain = supplyChains[0]
+const urls = urlBuilder(supplyChain);
+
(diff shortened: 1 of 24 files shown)


cypress/integration/sap_filters_spec.js
```diff
@@ -0,0 +1,96 @@
+import users from '../fixtures/user.json'
```
dev-digital, line 1: Tests for SAP filters are lighter side, particularly for OGD users.

With 2e2 tests using *mock-sso* to validate, there doesn't appear a way to change authenticated user easily. I tried to edit env vars within docker-compose.override.yaml to suit my need but it kept on replying with hard coded test user. I am sure this can be tweaked through but not within in PR :grin:

update_supply_chain_information/action_progress/templates/action_progress_dept.html
```diff
@@ -0,0 +1,9 @@
+{% extends "action_progress_base.html" %}
+
+{% block page_content %}
+
+<h3>DEPT</h3>
```
dev-digital, line 5: This will be removed within Listing(Next) work item

NickFitz (approved):
👍

PR #211, opened 2021-08-31
RT-489: SAU validation for updates must cover the whole period

dev-digital:
Bug fix monthly updates from admin panel


NickFitz (approved):
👍

PR #212, opened 2021-08-31
RT-484: Decrease the font size for notes within SAP details

dev-digital:
Only change to decrease the font of _notes_ fields to 16 pixels

### DIFF
--- a/update_supply_chain_information/action_progress/templates/includes/action_info.html
+++ b/update_supply_chain_information/action_progress/templates/includes/action_info.html
@@ -59,7 +59,7 @@
       </dd>
     </div>
     <div class="govuk-summary-list__row--no-border">
-        <dd class="govuk-summary-list__value" style="font-size: 16px;">
+        <dd class="govuk-summary-list__value govuk-!-font-size-16">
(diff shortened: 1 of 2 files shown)


update_supply_chain_information/action_progress/templates/includes/action_info.html
```diff
@@ -59,7 +59,7 @@ <h3 class="govuk-heading-s" style="padding-bottom: 10px;">Does this action affec
       </dd>
     </div>
     <div class="govuk-summary-list__row--no-border">
-        <dd class="govuk-summary-list__value">
+        <dd class="govuk-summary-list__value" style="font-size: 16px;">
```
NickFitz, line 62: Rather than using an inline style, you could use the class `govuk-body-s` which will also take care of adjusting the font size relative to other text on small screens: https://design-system.service.gov.uk/styles/typography/#body-small

dev-digital: Guess what, that's my first trial which didn't work. Hence to over-ride(to give higher precedence) I used in line styling.

NickFitz (approved):
👍

PR #214, opened 2021-09-01
RT-492: Include supply chain during SA lookup to guard against multiple objec…

dev-digital:
Include SC field during SA lookup to work around duplicated _slug_ field within SA model


NickFitz (approved):
👍

PR #218, opened 2021-09-03
RT-501: Using url_route to identify active menu instead of url_name to factor…

dev-digital:
Using _url_route_ to identify active menu instead of _url_name_ to factor in nested routes


NickFitz (approved):
:+1

PR #221, opened 2021-09-06
RT-486: Refactor SA.supporting_organisations field as charField

dev-digital:
**Changes**
- Change _supporting_organisations_ from _ChoiceField_ to _CharField_  
- Add supporting data migration to strip off existing special chars


NickFitz (approved):
👍

PR #229, opened 2021-09-14
RT-531: Make SupplyChain.risk_severity_status as optional field

dev-digital:
**Changes**
- Introduce _blank=True_ for SupplyChain.risk_severity_status field
- Tweak string based field to not use _null=True_ as per Django's recommendataion
- Test


NickFitz (approved):
:+1

PR #237, opened 2021-09-17
RT-541: Template tag to render GSC notes data.

NickFitz:
Just a template, tag, and tests.


update_supply_chain_information/chain_details/templates/gsc_notes.html
```diff
@@ -0,0 +1,6 @@
+{% if update_text %}
+    <p class="govuk-body">{{ update_text }}</p>
+{% endif %}
+{% if review_text %}
+    <p class="govuk-body">{{ review_text }}</p>
+{% endif %}
```
dev-digital, line 6: I am not clear on the requirements here.
Do we want to see 2 lines(html _p element_) for the gsc notes ?

update_supply_chain_information/chain_details/templatetags/gsc_notes.py
```diff
@@ -0,0 +1,25 @@
+from django import template
+from django.template.defaultfilters import date as date_filter
+
+register = template.Library()
+
+date_format = "l d M Y"
+
+
+@register.inclusion_tag("gsc_notes.html")
+def gsc_notes(instance):
```
dev-digital, line 10: Now that this tag being defined as _incluion_tag_, templates including this would call 

```
{% gsc_notes instance=instance %}
```

should we make sure _update_text_ is not pointing to empty string ? otherwise we could end up rendering empty space after the border/separator

update_supply_chain_information/chain_details/test/test_gsc_notes_templatetag.py
```diff
@@ -0,0 +1,117 @@
...
+
+@pytest.fixture(scope="session")
+def gsc_updated_on():
+    return date(year=2021, month=9, day=17)
+
+
+@pytest.fixture(scope="session")
+def gsc_review_on():
+    return date(year=2021, month=9, day=17)
+
```
dev-digital, line 23: I wonder if these fixtures could be replace with 3 lines of variable assignment :thinking:

dev-digital (commented):
Minor comments w/o approval

update_supply_chain_information/chain_details/templates/gsc_notes.html
NickFitz: As there's no design, and it seems like it might be unclear if it's all run on as one line, I assumed splitting the two sentences across two paragraphs would be best.

update_supply_chain_information/chain_details/test/test_gsc_notes_templatetag.py
NickFitz: Will do; I did them as fixtures while I was organising my thoughts about it, but turns out they don't really need to be.

update_supply_chain_information/chain_details/templatetags/gsc_notes.py
NickFitz: The template checks for an empty string:
```
{% if update_text %}
    <p class="govuk-body">{{ update_text }}</p>
{% endif %}
{% if review_text %}
    <p class="govuk-body">{{ review_text }}</p>
{% endif %}
```

NickFitz: In the case where neither is needed the template can be changed to something like:

```
{% if update_text or review_text %}
<div class="some-wrapper-we-find-out-we-need">
    {% if update_text %}
        <p class="govuk-body">{{ update_text }}</p>
    {% endif %}
    {% if review_text %}
        <p class="govuk-body">{{ review_text }}</p>
    {% endif %}
</div>
{%endif %}
```
but as that depends on how the markup ends up being implemented, I'm only including the basics with this PR.

dev-digital: We can talk about this in tomorrow's stand up, I guess. I am happy to approve as it is.

update_supply_chain_information/chain_details/test/test_gsc_notes_templatetag.py
dev-digital: Let me know once you fix these fixtures and I will approve

NickFitz: Already fixed in https://github.com/uktrade/update-supply-chain-information/pull/237/commits/3f70f3a9f4ae6b382975ff395e05f38e10a318d1

dev-digital: For some reason, I view was not being updated.

dev-digital (approved):

PR #238, opened 2021-09-17
RT-540: Abstract model that contributes GSC notes fields

NickFitz:
Model classes that inherit from `GSCUpdateModel` rather than `models.Model` will have the fields `gsc_last_changed_by`, `gsc_updated_on`, and `gsc_review_on`.


update_supply_chain_information/supply_chains/models.py
```diff
@@ -541,11 +541,28 @@ def __str__(self):
...
+    gsc_last_changed_by = models.TextField(
+        blank=True,
+        default="",
+        help_text="The entity responsible for the most recent change",
+    )
+    gsc_updated_on = models.DateField(
+        null=True, help_text="The date of the most recent change"
+    )
+    gsc_review_on = models.DateField(
+        null=True, help_text="The fdate when a review should be carried out"
```
dev-digital, line 554: minor type here

dev-digital (approved):
Minor typo spotted, looks good apart from it.

update_supply_chain_information/supply_chains/models.py
NickFitz: Just spotted that a few minutes ago myself :)

PR #327, opened 2022-07-20
Updating Django to 3.2.14 to resolve CVE-2022-34265

chris-pettinga:


rossmiller (approved):
LGTM!

holdenweb (approved):
Thanks for this. Do we have any policy about which version of Python to use, do you know?

chris-pettinga:
> Thanks for this. Do we have any policy about which version of Python to use, do you know?

Not sure if we have anything DIT-wide, this project uses 3.8 as defined in the Docker/Dockerfile.dev file however I wasn't able to spin up the image as I don't have the appropriate .env file. Figured the bug was too critical to wait around for it so I settled on using my system version (3.9) - hence the discrepancy.

chris-pettinga:
> > Thanks for this. Do we have any policy about which version of Python to use, do you know?
> 
> Not sure if we have anything DIT-wide, this project uses 3.8 as defined in the Docker/Dockerfile.dev file however I wasn't able to spin up the image as I don't have the appropriate .env file. Figured the bug was too critical to wait around for it so I settled on using my system version (3.9) - hence the discrepancy.

Ignore me I was being stupid, could've easily used the docker in hindsight - not sure I was thinking.

